PR #15, opened 2022-10-11
Random waveform injection improvements

EthanMarx:
1. Specifying `snr` is now `Optional`. If `None`, will not perform any snr reweighting
2. `forward` method now returns the injected tensor `X`, the `indices` of `X` where injections where performed, and the `sample_params` that correspond to those indices.
3.  Can now optionally pass `intrinsic_parameters`. This is a `Tensor` that contains the intrinsic parameter used to generate the polarizations passed. During the `forward` call, the parameters corresponding to the random injections will be appended to the sampled extrinsic parameters and returned.

### DIFF
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -27,7 +27,6 @@ class RandomWaveformInjection(torch.nn.Module):
         highpass: Optional[float] = None,
         prob: float = 1.0,
         trigger_offset: float = 0,
-        fixed: bool = False,
         **polarizations: np.ndarray,
     ) -> None:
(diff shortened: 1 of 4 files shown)


alecgunny (approved):
I think this is good to merge. Part of me wonders if the most basic, general implementation of this doesn't have SNR reweighting or intrinsic parameters at all, and that these can and should be implemented pretty straightforwardly via subclasses. But this is at least a good outline of all the potential features we would eventually want, and we can figure out how to break them down into separate objets later if users think the API is too complex.

EthanMarx:
> 

I completely agree. The core functionality is kernel sampling and projecting with random extrinsic parameters.

PR #21, opened 2022-11-02
fix SNR calculation in compute_snr_ifo, add test against lalsimulation

deepchatterjeeligo:
SNR for a (30, 30) system at 100 Mpc is computed with lalsimulation tools, and checked against that in `test_compute_ifo_snr` with a 10% relative tolerance


alecgunny (approved):
Looks great!

PR #22, opened 2022-11-03
Transform overhaul

alecgunny:
- Preferring buffers over parameters in `RandomWaveformInjection` for both internal consistency and consistency with Torch's preferred syntax for tensors that should be associated with a module (e.g. get moved around via `to`) but not tracked with gradients
- Making buffers for extrinsic and intrinsic parameters, as well as frequency mask, non-persistent, i.e. they don't get tracked with `RandomWaveformInjection.state_dict` and therefore don't get saved. Consistent with the model that a layer that has been fit on some background for some IFOs can be saved and then loaded again to project a totally new set of waveforms and their parameters.
- Initializing `RandomWaveformInjection.background` to zeros in `__init__` so that the PSD of a previously saved layer can be loaded back in to a new one without it having to call `.fit` (or even have access to the original data). In order to avoid having to check if `(background == 0).all()` on the forward call, introduced an internal `_has_fit` flag to keep track of whether `.fit` has been called.
- Polarizations are tracked via regular tensors, since these can have a large memory footprint and you don't necessarily want them living where all the other tensors live. They can either be moved manually, or by providing a `waveforms=True` kwarg to `RandomWaveformInjection.to`.
- A `device` kwarg has been added to `RandomWaveformInjection.sample` to control where the polarization tensors get moved to. It might make more sense to just move them automatically to `self.tensors.device`, but this felt like it allowed more flexibility. In `forward`, `X.device` is passed to this argument
- `window` and `scale` are saved as buffers in `SpectralDensity`. Doesn't make a big difference now, but could be useful once more exotic windows and scaling factors are allowed.

### DIFF
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -128,15 +128,14 @@ class RandomWaveformInjection(torch.nn.Module):
         self.trigger_offset = int(trigger_offset * sample_rate)
 
         # store ifo geometries
-        self.ifos = ifos
         tensors, vertices = gw.get_ifo_geometry(*ifos)
-        self.tensors = torch.nn.Parameter(tensors, requires_grad=False)
-        self.vertices = torch.nn.Parameter(vertices, requires_grad=False)
+        self.register_buffer("tensors", tensors)
+        self.register_buffer("vertices", vertices)
 
         # make sure we have the same number of waveforms
         # for all the different polarizations
         num_waveforms = waveform_size = None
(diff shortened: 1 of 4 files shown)


EthanMarx (approved):

alecgunny:
Confirmed this works. Using BBHNet as a benchmark, leaving the polarizations on CPU costs us ~4% on throughput, but we can make that up by going to larger batch sizes independent of the size of the waveform dataset that we use.

PR #24, opened 2022-11-03
refactor ifo snr test, add network snr test

deepchatterjeeligo:
fixes #21


alecgunny:
@deepchatterjeeligo could you add a test for `reweight_snrs` as well? Should probably look a lot like the `compute_network_snrs` tests, just with the waveforms reweighted to some target SNR.

alecgunny (approved):

Issue #26, opened 2022-11-06
Home for distributions module


EthanMarx:
The [distributions](https://github.com/ML4GW/BBHNet/blob/main/libs/data/bbhnet/data/distributions.py) used in BBHnet have now also been copied over to the PE project. It might be a good idea to create a module for them in ml4gw, especially considering their close connection with the `RandomWaveformInjection`  class.

PR #27, opened 2022-11-08
migrate distributions module

EthanMarx:
Move distributions module used in BBHnet and PE project


EthanMarx:
closes #26

alecgunny (approved):
Wow I have no idea how I missed this, I'm sorry about that. Looks great thanks for putting this together.

PR #29, opened 2022-12-06
Added error message to check for negative minimum values in slicing.py

wbenoit26:
Quick PR just to add an error message and fix some documentation in `slicing.py`


alecgunny:
Curious about why the diff is showing the whole file changing, wonder if it's due to line endings. Out of curiosity, what system did you implement the change on?

alecgunny:
Ok got it, can you really quickly run `file ml4gw/utils/slicing.py` and copy the output here?

wbenoit26:
Here you go: `ml4gw/utils/slicing.py: Python script, ASCII text executable, with CRLF line terminators`

alecgunny (approved):

alecgunny (changes requested):
On second thought, it might make sense to add a test that checks for this error specifically to ensure that users aren't stuck with that annoying lower-level Torch error. Why don't you add that and then wait until #31 merges to make sure the test passes and then this should be good to go.

wbenoit26:
What do you mean by that? Doesn't raising the `ValueError` avoid the Torch error?

alecgunny:
It does, we should just create a test (or a step in the existing test) that deliberately raises with error and catch it to make sure we get the error we want. Something like

```python
with pytest.raises(ValueError) as exc_info:
    sample_kernels(bad, combination, of, args)
assert str(exc_info.value).startswith("Kernel size")
```

alecgunny (approved):

PR #30, opened 2022-12-15
Fixed crlf issue

alecgunny:
Switching all line-endings to Unix style for consistency, and setting `.gitattributes` to enforce normalization going forward


alecgunny:
For those reviewing, click on the gear icon and set "Hide whitespace" to show that nothing other than `.gitattributes` has actually changed.

wbenoit26 (approved):

PR #31, opened 2022-12-15
Updating environment to fix issues and test pip install

alecgunny:


alecgunny:
Seems as if the issue that was breaking things was that the default versions of python shipping with the `setup-python@v2` workflow had started using `packaging>=22.0`, which has known issues with `poetry==1.2.0a2`. This fixes this by downgrading the packaging version only for this poetry version.

alecgunny:
Scratch that: the issue was the tox installs were updating packaging (poetry was rightly installing the version it needed). Moving the packaging downgrade until after this should (hopefully) fix this

wbenoit26 (approved):

PR #33, opened 2023-01-23
Whitening transform

### LINKED ISSUE #6: Add whitening transform from BBHNet
ML4GW/BBHNet has a whitening transform layer implemented [here](https://github.com/ML4GW/BBHNet/blob/main/libs/data/bbhnet/data/transforms/whitening.py). This should be moved into this repo and generalized.

alecgunny:
Closes #6 . Creating a whitening transform with an associated base class for transforms with parameters that need to be fit to some data (but not by gradient descent). Slightly reorganized from the implementation in BBHNet, with:
- Only arguments needed to set the size of the time domain filter are required at `__init__` time.
- Arguments needed to set the values of the time domain filter, which are arguments pertaining to how the background data should be converted to the frequency domain and then converted into a filter.

### DIFF
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -7,13 +7,14 @@ import torch
 
 from ml4gw import gw
 from ml4gw.spectral import Background, normalize_psd
+from ml4gw.transforms.transform import FittableTransform
 from ml4gw.utils.slicing import sample_kernels
 
 Distribution = CallableType[[int], np.ndarray]
 SourceParameter = Union[np.ndarray, Distribution]
 
 
-class RandomWaveformInjection(torch.nn.Module):
+class RandomWaveformInjection(FittableTransform):
     def __init__(
         self,
         sample_rate: float,
@@ -212,7 +213,7 @@ class RandomWaveformInjection(torch.nn.Module):
             self.register_buffer("background", buff)
         else:
             self.background = None
-        self._has_fit = False
+            self.built = True
 
     def to(self, device, waveforms: bool = False):
         super().to(device)
@@ -271,10 +272,11 @@ class RandomWaveformInjection(torch.nn.Module):
         if self.snr is None:
             raise TypeError("Cannot fit to backgrounds if snr is None")
 
-        sample_rate = sample_rate or self.sample_rate
         psds = []
         for ifo, background in backgrounds.items():
-            psd = normalize_psd(background, self.df, sample_rate, fftlength)
+            psd = normalize_psd(
+                background, self.df, self.sample_rate, sample_rate, fftlength
+            )
 
             # since this is presumably real data, there shouldn't be
             # any 0s in the PSD. Otherwise this will lead to NaN SNR
@@ -295,8 +297,7 @@ class RandomWaveformInjection(torch.nn.Module):
                 "Conversion to torch tensor pushed "
                 "background ASD values to 0"
             )
-        self.background.copy_(background)
-        self._has_fit = True
+        super().build(background=background)
 
     def _sample_source_param(
         self, param: SourceParameter, idx: gw.ScalarTensor, N: int
@@ -310,6 +311,13 @@ class RandomWaveformInjection(torch.nn.Module):
             return param(N).to(self.tensors.device)
         else:
             return param[idx]
+
+    def __call__(self, *args, **kwargs):
+        """
+        Override __call__ method to original Module call
+        since we do the `built` check in `sample`, not `forward`
+        """
+        return torch.nn.Module.__call__(self, *args, **kwargs)
 
     def sample(
         self,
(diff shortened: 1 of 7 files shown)


ml4gw/transforms/whitening.py
```diff
@@ -0,0 +1,253 @@
...
+        kernel_length = torch.zeros((1,))
+        self.register_buffer("kernel_length", kernel_length)
+
+        # set up a window that we won't save with state
+        # since it doesn't depend on the data at all
+        window = torch.zeros((self.ntaps,), dtype=dtype)
+        window[:-1] = torch.hann_window(self.ntaps - 1)
+        self.register_buffer("window", window, persistent=False)
+
+        self._has_fit = False
```
EthanMarx, line 120: this is unnecessary given the `FittableTransform` logic correct?

ml4gw/transforms/whitening.py
```diff
@@ -0,0 +1,253 @@
...
+        self._check_kernel_length(kernel_length)
+        df = 1 / kernel_length
+        ncorner = int(highpass / df) if highpass else 0
+
+        tdfs = np.zeros((self.num_channels, 1, self.ntaps - 1))
+        for i, (channel, x) in enumerate(backgrounds.items()):
+            psd = normalize_psd(
+                x, df, self.sample_rate, sample_rate, fftlength
+            )
+            if (psd == 0).any():
```
EthanMarx, line 200: See above comment. Maybe in the future `normalize_psd` can handle the logic of rescaling data so that we don't run into this precision issue.

tests/transforms/test_whitening.py
```diff
@@ -0,0 +1,142 @@
...
+
+            # TODO: this isn't the strongest check in the world,
+            # but it's the closest I can get for now. Check that
+            # at least 95% of the outputs are within 1% of the
+            # value produced by gwpy
+            close = np.isclose(result, ts, rtol=1e-2)
+            assert close.mean() > 0.95
+
+
+def test_whitening_save_and_load(dtype, tmp_path):
```
EthanMarx, line 116: What would happen if you attempt to save and load before calling `.fit` ? Is the idea here that regardless, the user will be saved by the `self._check_built` at call time ?

ml4gw/transforms/injection.py
```diff
@@ -272,47 +272,23 @@ def fit(
...
+                background, self.df, self.sample_rate, sample_rate, fftlength
+            )
 
             # since this is presumably real data, there shouldn't be
             # any 0s in the PSD. Otherwise this will lead to NaN SNR
             # values at reweighting time. TODO: is this really a
             # constraint we want to enforce, or should we leave this
             # to the user to catch?
-            if (background == 0).any():
+            if (psd == 0).any():
```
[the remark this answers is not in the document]
alecgunny: Yeah I think I agree here, I'll remove this

ml4gw/transforms/injection.py
```diff
@@ -455,6 +432,17 @@ def forward(
...
+
+            if N == 0:
+                # return empty parameters if our roll
+                # of the dice didn't turn up any waveforms
+                param_shape = 3
+                if self.snr is not None:
+                    param_shape += 1
+                if self.intrinsic_parameters is not None:
+                    param_shape += self.intrinsic_parameters.size(-1)
+                return X, torch.zeros((0,)), torch.zeros((0, param_shape))
```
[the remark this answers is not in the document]
alecgunny: I'm not sure it's necessarily ambiguous: the tensors just represent the information about the waveforms that got sampled, which in this case was none of them (spent like 30 seconds resisting typing that word...). I guess the logic here is that by returning the tensors, you can carry on with the typical logic ~90% of the time without having to insert special checks (e.g. you can set `y[idx] = 1`, you can do algebraic manipulations to `params`, etc.), and for the remaining use cases checking `if len(idx) == 0` is as intrusive as checking `if idx is None`

ml4gw/transforms/transform.py
```diff
@@ -0,0 +1,33 @@
...
+    def __call__(self, *args, **kwargs):
+        self._check_built()
+        return super().__call__(*args, **kwargs)
+
+    def load_state_dict(
+        self, state_dict: Mapping[str, Any], strict: bool = True
+    ):
+        keys = super().load_state_dict(state_dict, strict)
+        self.built = True
+        return keys
```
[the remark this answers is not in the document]
alecgunny: Hmm that's an interesting question. My first instinct is that if you want to save a parameter of all 0s that's your business, but on the other hand the reality is that that will often be the result of having made a mistake. Maybe a warning is a good middle ground?

ml4gw/transforms/injection.py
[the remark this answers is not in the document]
EthanMarx: Okay agreed. I tricked myself by reading the `torch.zeros` as returning arrays of 0s but the empty arrays makes perfect sense.

tests/transforms/test_whitening.py
alecgunny: See above. Right now, we wouldn't catch it, but we can only do so much to save users from themselves

EthanMarx: Fair point, should be fine for now then. Can always add more guard rails if it becomes a persistent problem

EthanMarx (approved):

PR #41, opened 2023-01-29
Wrote unfolding function

wbenoit26:
Threw this together based on the code in `spectral.py`. Is this all the function should do, or would we want more flexibility than this? The `Unfold` op confuses me, so I'm not seeing exactly how this would be used.


alecgunny:
This is a really good start! Couple things I would suggest:
- This could probably live in `utils/slicing.py` since it's a helper op for slicing windows out of data (in fact if you called `slice_kernels` with evenly spaced `idx` you'd get something basically like this, albeit slower)
- I would recommend using [`torch.nn.functional.unfold`](https://pytorch.org/docs/stable/generated/torch.nn.functional.unfold.html) rather than constructing the op then calling it on the tensor. Not a big difference, but I think restricting the use of modules to when you're stacking layers in a network is a good distinction to make.
- After unfolding, we should reshape to whatever's sensible given the input shape (note that I do that after doing last axis operation in `spectral.py` [here](https://github.com/ML4GW/ml4gw/blob/443bd20a70c56dad7c3e3732dd7f79bf9e88d762/ml4gw/spectral.py#L378):
    - For 1D, return a 2D tensor `(num_windows, kernel_size)`
    - For 2D, the assumption is that this a multichannel timeseries (not a batch of single channel timeseries. This is a convention I think we enforce pretty universally here, and I think is fair given that if you want the latter you can always add a dummy channel dimension). So we should return a 3D tensor, whose dims I _think_ we want to be ordered as `(num_windows, num_channels, kernel_size)`, since slicing along the 0th axis would give us tensors like the original input, and would allow us to do algebra with `(num_channels, kernel_size)` shaped tensors (e.g. if `other` is a tensor with shape `(num_channels, kernel_size)`, doing `(windowed * other).sum((-2, -1))` would effectively be like doing a convolution along each channel with stride `stride`).
    - For 3D, return a 4D tensor with shape `(num_windows, batch_size, num_channels, kernel_size)`
- Could be worth adding a boolean arg `drop_last: bool = True` that, if the timeseries can't be evenly broken up into kernels, decides whether the remainder at the end gets returned or not. I can foresee this being helpful for making a tensor-friendly version @EthanMarx 's overlap-add whitening implementation, but this also might be adding unnecessary complexity.

wbenoit26:
@alecgunny is this the kind of thing you're looking for with the remainder? I should probably add a check for whether the kernel size evenly divides the time series so that no remainder is returned in that case, but this was my interpretation of what that should look like.

ml4gw/utils/slicing.py
```diff
@@ -13,6 +14,67 @@
...
+    if x.ndim == 1:
+        x = x[None, None, None, :]
+    elif x.ndim == 2:
+        x = x[None, :, None, :]
+    elif x.ndim == 3:
+        x = x[:, :, None, :]
+
+    num_windows = (x.shape[-1] - window_size) // stride + 1
+    stop = (num_windows - 1) * stride + window_size
+    x_clone = x.detach().clone()
```
alecgunny, line 61: Is there a reason you need to detach from the graph and copy here?

ml4gw/utils/slicing.py
```diff
@@ -13,6 +14,67 @@
...
+    """
+    reshape = list(x.shape[:-1])
+    if x.ndim == 1:
+        x = x[None, None, None, :]
+    elif x.ndim == 2:
+        x = x[None, :, None, :]
+    elif x.ndim == 3:
+        x = x[:, :, None, :]
+
+    num_windows = (x.shape[-1] - window_size) // stride + 1
```
alecgunny, line 59: I think you can do this calc before the reshape, since the last axis doesn't change, and can slice out the remainder here while it already has the shape that you want. Something like:

```python
num_windows = (x.shape[-1] - window_size) // stride + 1
remainder = x.shape[-1] - window_size - num_windows * stride
if remainder == 0:
    if not drop_last:
        # create an empty tensor for consistency
        shape = list(x.shape)
        shape[-1] = 0
        remainder = torch.zeros(shape, dtype=x.dtype, device=x.device)
else:
    # separate x from its remainder _regardless_ of whether
    # we end up returning it or not
    x, remainder = torch.split(x, [x.shape[-1] - remainder, remainder], dim=-1)
```
Then at the end you can just do
```python
if not drop_last:
    return x, remainder
return x
```

alecgunny (commented):
Couple questions/comments, then I think just needs tests and should be good to go. For testing all I think you need to check is that e.g.

```python
unfold_windows([1, 2, 3, 4, 5, 6], window_size=3, stride=2) == [[1, 2, 3], [3, 4, 5]]]
unfold_windows(
    [[1, 2, 3, 4, 5, 6], [2, 3, 4, 5, 6, 7]], window_size=3, stride=2
) == [[[1, 2, 3], [2, 3, 4]], [[3, 4, 5], [4, 5, 6]]]

# ... etc. for 3D, then verify drop last functionality, e.g. for
# the first example above the second item returned should
# just be [6], for the second it should be [[6], [7]], etc.
```

ml4gw/utils/slicing.py
wbenoit26: Honestly, I can't remember what I was thinking last week. I _think_ I wanted to have only one `if not drop_last` block, so I needed to copy `x` before it underwent its transformations, and I read that `.detach().clone()` was the preferred was to make a copy of a tensor. But that's not really necessary

ml4gw/utils/slicing.py
wbenoit26: So we want `remainder` to have the same dimensionality as the input tensor, not the output? Or should I still reshape it before returning?

alecgunny: Hmmm that's a good question. I was thinking yes yesterday, but maybe it does make sense to prepend a dummy dimension for the "one window" this constitutes. I guess it depends on what the most common use cases for the last bit are, and that will come with time. Probably not worth agonizing over so I'll leave it up to you, whatever you think is more consistent I'm cool with

ml4gw/utils/slicing.py
alecgunny: Ahhh ok gotcha. Yeah I think in this case you would just want to change the variable name for `x` once you started transforming it, because when you `detach` a tensor you stop all backwards gradient flow through that tensor (it emerges from the copy as if it was just born, with no relation to any transforms that might have brought the tensor from which it was cloned into being. If you've seen The Prestige, it's kind of like that). This means that things in the loss function which depend on this chunk won't be able to update the weights of transforms before it accordingly.

wbenoit26:
@alecgunny Do you know why the `poetry-lock` pre-commit check is failing here? It passes on my local version

ml4gw/utils/slicing.py
```diff
@@ -13,6 +14,76 @@
...
+
+    x = unfold(x, (1, num_windows), dilation=(1, stride))
+    reshape += [num_windows, -1]
+    x = x.reshape(*reshape)
+    x = x.transpose(1, -2).transpose(0, 1)
+
+    if not drop_last:
+        reshape[-2] = 1
+        remainder = remainder.reshape(*reshape)
+        remainder = remainder.transpose(1, -2).transpose(0, 1)
```
alecgunny, line 82: I think all you should need to do here is `remainder = remainder[None]` to add a dummy 1st dimension, since the rest of the dims are what we want, right?

alecgunny: (or 0th dimension I guess)

wbenoit26: Right, yes, I forgot to move the reshape

alecgunny (approved):

PR #44, opened 2023-02-23
Match test set snr

wbenoit26:
First pass at some minimally invasive changes that allow `target_snrs` to be a shuffled version of `snrs`. The idea would just be to set `snr=-1` [here](https://github.com/ML4GW/BBHNet/blob/3e9150cb8bbf5a10e71ac8fe2aae9a3d7ad8949d/projects/sandbox/train/train/utils.py#L120).

I had trouble coming up with a less awkward way to do this that didn't involve rewriting a lot of things. It makes sense to get the `target_snrs` prior to calculating the signal SNRs so that we guarantee that the SNRs stay coupled to the ifo responses that generated them, but that means that there needs to be a way to indicate what we want before the signal SNRs exist. I thought about some kind of identity distribution, but I didn't come up with anything that made sense.

- [x] Add tests
- [x] Less clunky implementation

### DIFF
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -175,7 +175,7 @@ class RandomWaveformInjection(FittableTransform):
         names = ["dec", "psi", "phi", "snr"]
         for name, param in zip(names, [dec, psi, phi, snr]):
             if name == "snr" and isinstance(param, int):
-                self.register_buffer(name, param, persistent=False)
+                self.snr = -1
             elif not isinstance(param, Callable) and param is not None:
                 try:
                     length = len(param)
@@ -388,7 +388,7 @@ class RandomWaveformInjection(FittableTransform):
             snrs = gw.compute_network_snr(
                 ifo_responses, self.background, self.sample_rate, self.mask
             )
-            if isinstance(self.snr, int) and self.snr == 1:
+            if isinstance(self.snr, int) and self.snr == -1:
                 idx_perm = torch.randperm(N)
                 target_snrs = snrs[idx_perm]
             else:
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -174,7 +174,9 @@ class RandomWaveformInjection(FittableTransform):
         # number of waveforms
         names = ["dec", "psi", "phi", "snr"]
         for name, param in zip(names, [dec, psi, phi, snr]):
-            if not isinstance(param, Callable) and param is not None:
+            if name == "snr" and isinstance(param, int):
+                self.register_buffer(name, param, persistent=False)
+            elif not isinstance(param, Callable) and param is not None:
                 try:
                     length = len(param)
                 except AttributeError:
@@ -386,7 +388,7 @@ class RandomWaveformInjection(FittableTransform):
             snrs = gw.compute_network_snr(
                 ifo_responses, self.background, self.sample_rate, self.mask
             )
-            if (self.snr == -1).all():
+            if isinstance(self.snr, int) and self.snr == 1:
                 idx_perm = torch.randperm(N)
                 target_snrs = snrs[idx_perm]
             else:
--- a/ml4gw/transforms/injection.py
+++ b/ml4gw/transforms/injection.py
@@ -383,10 +383,10 @@ class RandomWaveformInjection(FittableTransform):
         )
 
         if self.snr is not None:
-            snrs = gw.compute_network_snrs(
+            snrs = gw.compute_network_snr(
                 ifo_responses, self.background, self.sample_rate, self.mask
             )
-            if self.snr == -1:
+            if (self.snr == -1).all():
                 idx_perm = torch.randperm(N)
                 target_snrs = snrs[idx_perm]
             else:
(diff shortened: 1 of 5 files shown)


alecgunny:
I actually don't know that I think your implementation is too clunky. The one thing I might change if we went this direction would be that
```python
if snrs.dim() == 0:
    return responses
```
Otherwise, from a more conceptual standpoint, my only hang up is that it feels like the `gw` submodule is meant mostly as a set of utility libraries distinct from the particular use case they're being applied to. So adding stochastic transformations, which are useful for NN training but probably not really anywhere else, might be cramming too much functionality in here.

You're right that the bummer is that we can't pass `target_snrs` without having to duplicate the compute of computing the snrs to begin with, but the argument would be that then this higher-level utility isn't for you and you'll have to go one level deeper, which isn't too bad in this case since the underlying logic is super simple. So some alternatives might be:

- Handle this at the application (i.e. BBHNet) level: compute all the SNRs up front and pass this tensor as the `snr` argument of the constructor. Probably the simplest, and maybe slightly more robust since the possible SNRs aren't limited to those in your batch, but won't be sufficient in the longer term as we move towards out-of-memory dataloading.
- Replicate the underlying logic inside `RandomWaveformGenerator.sample`:
```python
if self.snr is not None:
    snrs = gw.compute_network_snrs(ifo_responses, self.background, self.sample_rate, self.mask)
    if self.snr == -1:
        idx = torch.randperm(N)
        target_snrs = snrs[idx]
    else:
        target_snrs = self._sample_source_param(self.snr, idx, N)

    weights = target_snrs / snrs
    rescaled_responses = ifo_responses * weights.view(-1, 1, 1)
```
Since I think this is the only place that actually uses the SNR rescaling at the moment, there's an argument that this would imply we should just can the function altogether since it's really just a couple lines of code anyway, but that probably wouldn't be worth holding up this PR to think about.

In any case, just some possibilities to think about. We can discuss more at the BBHNet sync today too.

wbenoit26:
Didn't see this comment before I changed things around, but it gives me some things to think about

ml4gw/transforms/injection.py
```diff
@@ -382,14 +383,17 @@ def sample(
...
-            rescaled_responses = gw.reweight_snrs(
-                ifo_responses,
-                target_snrs,
-                backgrounds=self.background,
-                sample_rate=self.sample_rate,
-                highpass=self.mask,
+            snrs = gw.compute_network_snr(
+                ifo_responses, self.background, self.sample_rate, self.mask
             )
+            if (self.snr == -1).all():
```
alecgunny, line 389: So is the idea here to pass a tensor of all `-1`s at `__init__` time? It might be more intuitive to just be able to pass a literal `-1`. I know that torch is finnicky about comparing tensors to other stuff, so it might be easiest to include a type check here: `if isinstance(self.snr, int) and self.snr == -1:` (though I guess since `-1` is the only `int` we allow this is probably redundant)

alecgunny: I think we'll want to do a special case of `if name == "snr" and isinstance(param, int):` during `__init__`. Doing explicit `int` checking isn't necessarily the most recommended pythonic behavior, but it's easier than adding in a separate catch for `TypeError` during the `len` check at line 179

alecgunny (changes requested):
Just implementation questions about how we indicate the -1 to the constructor

ml4gw/transforms/injection.py
wbenoit26: Yeah, this was to account for the option that `self.snr` was an array of SNRs. The type checking makes more sense though

wbenoit26: Am I right in saying that we should have `SourceParameter = Union[np.ndarray, torch.Tensor, Distribution]`? Just because of how we use it in `_sample_source_param`.

ml4gw/transforms/injection.py
```diff
@@ -173,7 +174,9 @@ def __init__(
         # number of waveforms
         names = ["dec", "psi", "phi", "snr"]
         for name, param in zip(names, [dec, psi, phi, snr]):
-            if not isinstance(param, Callable) and param is not None:
+            if name == "snr" and isinstance(param, int):
+                self.register_buffer(name, param, persistent=False)
```
alecgunny, line 178: You don't need to register it as a torch parameter, I don't think it will type check as an `int` in this case

ml4gw/transforms/injection.py
```diff
@@ -382,14 +385,17 @@ def sample(
...
-            rescaled_responses = gw.reweight_snrs(
-                ifo_responses,
-                target_snrs,
-                backgrounds=self.background,
-                sample_rate=self.sample_rate,
-                highpass=self.mask,
+            snrs = gw.compute_network_snr(
+                ifo_responses, self.background, self.sample_rate, self.mask
             )
+            if isinstance(self.snr, int) and self.snr == 1:
```
alecgunny, line 391: Wait do we want `1` or `-1`?

alecgunny (commented):
I think all we would need would just be to set `self.snr = -1` and then just check for this at call time

ml4gw/transforms/injection.py
wbenoit26: Dang it, oops, good catch

ml4gw/transforms/injection.py
wbenoit26: Ahhh, okay, I was getting my head turned around with all the types

alecgunny (approved):

PR #47, opened 2023-04-09
`InMemoryDataset` inherit from `torch.utils.data.Dataset`

EthanMarx:
`InMemoryDataset` now inherits from `torch.utils.data.Dataset` and implements `__getitem__`

Motivation:
Been playing around with the `torch.distributed` library. From what I can tell, the easiest way to get this going is to wrap a `torch.utils.data.Dataset` in a `torch.utils.data.distributed.DistributedSampler`.

### DIFF
--- a/ml4gw/dataloading/in_memory_dataset.py
+++ b/ml4gw/dataloading/in_memory_dataset.py
@@ -3,11 +3,12 @@ from typing import Optional, Tuple, Union
 
 import numpy as np
 import torch
+from torch.utils.data import Dataset
 
 from ml4gw.utils.slicing import slice_kernels
 
 
-class InMemoryDataset:
+class InMemoryDataset(Dataset):
     """Dataset for iterating through in-memory multi-channel timeseries
 
     Dataset for arrays of timeseries data which can be stored
@@ -207,6 +208,22 @@ class InMemoryDataset:
         self._idx = idx
         self._i = 0
         return self
+
+    def __getitem__(
+        self, idx: torch.Tensor
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+        # slice the array of _indices_ we'll be using to
+        # slice our timeseries, and scale them by the stride
+        idx = self._idx[idx] * self.stride
+
+        # slice our timeseries
+        X = slice_kernels(self.X, idx, self.kernel_size)
+        if self.y is not None:
+            y = slice_kernels(self.y, idx, self.kernel_size)
+
+        if self.y is not None:
+            return X, y
+        return X
 
     def __next__(
         self,
--- a/ml4gw/dataloading/in_memory_dataset.py
+++ b/ml4gw/dataloading/in_memory_dataset.py
@@ -218,12 +218,11 @@ class InMemoryDataset(Dataset):
 
         # slice our timeseries
         X = slice_kernels(self.X, idx, self.kernel_size)
+        y = None
         if self.y is not None:
             y = slice_kernels(self.y, idx, self.kernel_size)
 
-        if self.y is not None:
-            return X, y
-        return X
+        return X, y
 
     def __next__(
         self,
@@ -243,13 +242,7 @@ class InMemoryDataset(Dataset):
         # slice the array of _indices_ we'll be using to
         # slice our timeseries, and scale them by the stride
         slc = slice(self._i * self.batch_size, (self._i + 1) * self.batch_size)
-        idx = self._idx[slc] * self.stride
-
-        # slice our timeseries
-        X = slice_kernels(self.X, idx, self.kernel_size)
-        if self.y is not None:
-            y = slice_kernels(self.y, idx, self.kernel_size)
-
+        X, y = self.__getitem__(slc)
         self._i += 1
         if self.y is not None:
             return X, y


wbenoit26:
That looks pretty cool, seems like a good idea. Should probably point this PR against `dev` rather than `main` (and also we should merge `dev` into `main` at some point and bump the version).

alecgunny:
I agree we should start inheriting from/using the built-in Torch APIs more, but I think there's a couple things to consider here:
- You can have a dataset that just implements the `__iter__` and `__next__` methods, see the distinction [here](https://pytorch.org/docs/stable/data.html#dataset-types) between iterable and map-style datasets. You would just need to inherit from [`IterableDataset`](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset).
- I think the risk with implementing `__getitem__` is that 1) if you implement both `__getitem__` and `__iter__`, I don't know which type of dataset Torch assumes this represents, 2) The _default_ behavior with using this with a `DataLoader` is that at iteration time, a batch of indices are sampled, `__getitem__` is called for each one of these indices, and the results are concatenated. I think calling `__getitem__` with just an integer index could potentially break `slice_kernels`, and even if not would defeat its purpose of efficiently slicing a batch of kernels. The iterative slice-then-concatenate would probably be much slower than what we have now. You can, however, get around this with a custom sampler object, that would e.g. pass batches of indices, which would have the benefit of separating out some of that index shuffling logic that we call during `__iter__`
- Most of the Torch dataloading APIs assume that your data is in the form of numpy arrays, which are converted to Torch tensors at collation time (i.e. once either the `__next__` or `__getitem__`s have been called). I don't think it'll necessarily be a problem to have these methods return torch Tensors, which we'd want to do since all of our fancy slicing is done in Torch (though maybe replicating these with numpy, which would actually probably be simpler, at some point isn't a bad idea either). I definitely envision there being unexpected problems, however, when these Tensors are on GPU already, especially if you want to take advantage of multiprocessed or distributed dataloading. With that in mind, it probably makes sense to strip out the pre-mapping to device in `__init__` and let people call `.to` themselves on the returned tensors. I think we're using large enough datasets now in most our use cases where this is more sensible behavior anyway, and we're using large enough networks that if you use multiprocessing and pinned memory you should be able to fully parallelize this with the optimization step and not cost yourself anything but the transfer itself (TensorFlow's data API has a nice option to prefetch tensors to GPU asynchronously, which it's a bummer Torch doesn't have)

EthanMarx:
Thanks for the feedback. 

- Didn't realize this existed. Removes the need for `__getitem__` which will eliminate the ambiguity you describe in your second bullet
- I think if we inherit from `IterableDataset` and set `batch_size=None`and `batch_sampler=None` in the `DataLoader` (i.e. [disable automatic batching](https://pytorch.org/docs/stable/data.html#disable-automatic-batching)), the functionality of `InMemoryDataset` will remain the same.
- Can you elaborate what the problems may be if the tensors are already on GPU? In the process of iterating, will torch send them to cpu / expect them to be on cpu as numpy arrays? 

Anyways, thought i could whip something up that would speed up the PE training but alas the problems more complicated than I anticipated.

alecgunny:
So in general for using multiprocessing with GPU libs, the thing to keep in mind is that the objects that we manipulate in Python-land aren't the real tensors, they're just symbolic pointers to objects that live in a much lower-level CUDA/C++ land. Their memory and the execution of operations on them isn't handled by Python, but by Torch's C++ code that we attempt to control at our higher level. This is why you can't add multiprocessing to a Torch script and squeeze out extra performance: the actual execution of computational graphs is handled at a lower level, so while you could potentially add operations to that graph asynchronously, you don't actual impact the rate or manner with which Torch executes it. Torch also doesn't really like when you pass tensors between processes, that's why it has its own [multiprocessing submodule](https://pytorch.org/docs/stable/multiprocessing.html).

That said, taking a look more at this use case, it looks like it could actually work as long as the dataset object (and hence its underlying tensor) get initialized in each child subprocess, _not_ in the main process and then passed to each child. So as long as you made sure that the `rank` of the child process gets passed to `__init__` as the intended device, I think this might actually work ok. You wouldn't want to use `num_workers>0`, but looking at a few examples it looks like this isn't recommended for the distributed setting anyway (and we wouldn't need it since our data's already in-memory).

Having read these examples more, I see why `__getitem__` makes sense here: you pass a `DistributedSampler` to `DataLoader`, and that implicitly iterates through dataset indices and calls `__getitem__` on them. However, I actually don't know that we need to use the `DistributedSampler` here because our sense of an "epoch" is so nebulous, at least for use cases where we don't sample coincidentally. The odds of the _exact_ same sample appearing twice are extremely low (1 / N^2), and we sample way fewer than that in a given epoch such that I don't really see this as an issue (there's even some debate about how helpful the concept of an epoch really is. It's mostly a matter of how frequently you validate, which is a trade-off of trying to not overfit to the validation set vs. getting a more fine-grained sense of when you've "optimized" your model).

Where this _might_ make a difference is for waveforms, which could make sense to implement as a `Dataset` object. In the BBHNet case, there's no fixed batch size however, so you wouldn't use it with a `DataLoader` and instead would use the `DistributedSampler` directly to sample a batch of waveforms. However, in your PE case, it seems like you would have a fixed batch size, so you could imagine doing something like
```python
background_dataset = InMemoryDataset(..., device=f"cuda:{rank}")

# since your not actually using the sampler, I don't
# think you really need this, but including it for
# posterity
background_dataloader = DataLoader(background_dataset, sampler=None)

waveform_projector = NewWaveformProjectorClass(...)
waveform_dataset = NewWaveformDatasetClass(waveform_filename, preprocessor=waveform_projector)
waveform_sampler = DistributedSampler(waveform_dataset, rank)
waveform_dataloader = DataLoader(waveform_dataset, batch_size=batch_size, sampler=waveform_sampler)
for background, (waveform, params) in zip(background_dataloader, waveform_dataloader):
    waveform = waveform.to(background.device)
    injected = background + waveform
    ... do rest of training
```

PR #48, opened 2023-04-18
add log uniform distribution

EthanMarx:


ml4gw/distributions.py
```diff
@@ -73,6 +73,20 @@ def __call__(self, N: int) -> torch.Tensor:
...
+    Sample from a log uniform distribution
+    """
+
+    def __init__(self, low: float, high: float) -> None:
+        self.low = math.log(low)
+        self.high = math.log(high)
+
+    def __call__(self, N: int) -> torch.Tensor:
+        u = super().__call__(N)
+        return 10**u
```
alecgunny, line 87: `math.log` is the natural logarithm, so don't think this adds up. I'm fine with either one

ml4gw/distributions.py
```diff
@@ -73,6 +73,20 @@ def __call__(self, N: int) -> torch.Tensor:
...
+    """
+    Sample from a log uniform distribution
+    """
+
+    def __init__(self, low: float, high: float) -> None:
+        self.low = math.log(low)
+        self.high = math.log(high)
+
+    def __call__(self, N: int) -> torch.Tensor:
+        u = super().__call__(N)
```
alecgunny, line 86: If you're gonna take advantage of this as a child class, might be worth considering using `super.__init__(math.log(low), math.log(high))` above to ensure total consistency

alecgunny (commented):
Only major thing is the disparity between `math.log` and `10**`, which maybe I'm misunderstanding

alecgunny (approved):

PR #49, opened 2023-04-24
Sine gaussian waveform

EthanMarx:
Adds a `waveforms` module that contains a torch implementation of the `lalsimulation` sine-Gaussian. https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/lib/LALSimBurst.c#L1080

Adds a `types` module


ml4gw/distributions.py
```diff
@@ -108,3 +108,15 @@ def __call__(self, N: int) -> torch.Tensor:
         samples = self.x_min ** (-self.alpha + 1) - u
         samples = torch.pow(samples, -1.0 / (self.alpha - 1))
         return samples
+
+
+class ParameterSampler(torch.nn.Module):
```
alecgunny, line 113: Not sure this makes sense to implement here, since it's a higher-level API leveraging these more fundamental `Callable`s. Think it would make more sense to implement in the waveform submodule itself, or if we think this is something we'll want to use more generally creating a `sampling` submodule or something and having this and `distributions.py` live there. My inclination is to go with the former until we have a more general use case, but up for debate

ml4gw/waveforms/sine_gaussian.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __post_init__(self):
+        # determine times based on requested duration and sample rate
+        # and shift so that the waveform is centered at t=0
+        num = int(self.duration * self.sample_rate)
+        times = torch.arange(num, dtype=torch.float64) / self.sample_rate
+        times -= self.duration / 2.0
+
+        self.times = times
+        self.window = tukey_window(num)
```
alecgunny, line 34: My general take is that if you don't necessarily need the dataclass attributes in the class itself, but just to initialize some other properties it depends on, it's probably better form to use a more traditional `__init__`. The main benefit of the `dataclass` is to avoid the boilerplate of `__init__`s that just assign their arguments to attributes and don't do much else. I'm a frequent abuser of this as well so not a big deal but this definitely feels like more than a container class for some data fields

pyproject.toml
```diff
@@ -25,6 +25,8 @@ pytest = "^7.0"
 
 # need lalsuite to compute gmst in injection testing
 lalsuite = "^7.0"
+# useful for creating jupyter notebooks
+ipykernel = "^6.22.0"
```
alecgunny, line 29: Probably don't want to make this a general requirement of the lib. Possibly as a dev requirement?

alecgunny (commented):
Couple notes on structure but looks really nice

ml4gw/waveforms/sine_gaussian.py
EthanMarx: Yeah bad habit

pyproject.toml
alecgunny: Ah ok missed that, that makes sense then. The one thing I might suggest is to make this `jupyter` instead since that will install `ipykernel` by default, and then folks who are on LDG can use that to install a kernel and folks who are developing locally can run their own local server

ml4gw/waveforms/generator.py
```diff
@@ -0,0 +1,39 @@
+from typing import Callable
```
alecgunny, line 1: Part of me is wondering, given how simple this module is, if it makes sense to relegate this to downstream applications to implement as they see fit. Like to take BBHNet as an example, is using this simpler (or more clear) than going:

```
dec = self.dec(N)
psi = self.psi(N)
phi = self.phi(N)

idx = torch.randint(however_we_keep_track_of_num_waveforms, size=(N,))
polarizations = {k: v[idx] for k, v in self.polarizations.items()}
responses = self.projector(dec, psi, phi, **polarizations)
```

I do like the idea of the `ParameterSampler` as a sort of torch-ified extension of the bilby `PriorDict`, possibly adding `.prob` functionality down the line, but in that case it might make more sense as a `dict` subclass with a `.sample` method. The main benefit of torch `Module`s is for the tracking of parameters and the ability to export graphs of modules in one fell swoop, and I don't know that we need either of those benefits here.

EthanMarx: Yeah I agree both the `WaveformSampler` and `ParameterSampler` are quite thin wrappers at the moment. 
My vision was also an eventual bilby `PriorDict` like object with `.probs` etc. 

I'm really not sure what's best. I'm constantly debating what could benefit from generalization.

EthanMarx: I put together a [gist](https://gist.github.com/EthanMarx/919de0d738319581d0b97af42638f7ae) of what the PE projects injector would look like under the current implementation of the API to get a better sense of all the use cases.
 
I think one pro of the `ParameterSampler` is that it can abstract the sampling of many parameters in one object. This makes the PEInjector object agnostic to the waveform / parameters of the waveform. 

Then again, the `ParameterSampler` is something that could be implemented by downstream projects really easily as well... 

For the sake of getting the `SineGaussian` merged in happy to remove this module and continue this discussion in the injection refactoring PR.

alecgunny (approved):

Issue #20, opened 2022-11-01
`RandomWaveformInjection` refactoring


alecgunny:
As discussed under #15 , there are probably better ways to simplify and refactor `RandomWaveformInjection` to strip out its core functionality (projection with random extrinsic parameters) and then add more complicated functionality (SNR reweighting and returning of intrinsic parameters) as subclasses either in the library itself or for individual use cases to implement, with helper methods to make adding such functionality simpler (e.g. modularizing the index sampling so new `forward` methods require very little code)

alecgunny:
Given the additional functionality of #44 , it's probably time to bump this. I think the key will be to separate out the sampling of the extrinsic parameters, from the sampling of the waveforms/intrinsic parameters, from the projection, from the rescaling, from the injection. If we construct everything as `Module`s, they'll be really easy to string together. For the random sampling case, this could look like:

```python
# probably worth making the distributions `torch.nn.Module`s as well
# and starting to use `device` in the `torch.rand` calls to avoid transfer
class ParameterSampler(torch.nn.Module):
    def __init__(self, *parameters: Callable):
        ...

    def forward(self, N: int, device: Optional[str] = None):
        # optionally make it **parameters above and have this return a dict
        return (p(N, device) for p in self.parameters)


class WaveformSampler(torch.nn.Module):
    """
    Module _just_ for doing random sampling of waveforms
    and optionally any intrinsic parameters, saved as buffers
    """
    def __init__(self, paramters: Optional[torch.Tensor] = None, **polarizations: torch.Tensor):
        ...

    def foward(self, N: int):
        idx = torch.randint(self.num_waveforms, size=(N,))
        waveforms = {k: v[idx] for k, v in self.polarizations}
        if self.parameters is not None:
            return waveforms, self.parameters[idx]
        return waveforms


class WaveformProjector(torch.nn.Module):
    def __init__(*ifos: str):
        self.tensors = ...
        self.vertices = ...

    def forward(dec, phi, psi, **polarizations):
        ...


class SNRRescaler(FittableTransform):
    def __init__(self, distribution: Optional[Callable] = None, highpass, ...):
        self.distribution = distribution
        self.register_buffer("background", ...)

    def fit(....):
        ...

    def forward(self, waveforms):
        snrs = compute_network_snr(waveforms, self.background, ...)
        if self.distribution is None:
            idx = torch.randperm(len(snrs))
            target_snrs = snrs[idx]
        else:
[...]

PR #50, opened 2023-04-24
Injection refactoring

EthanMarx:
Refactoring of `RandomWaveformInjection` object


EthanMarx:
@wbenoit26 @alecgunny I've essentially implemented what has been outlined in #20. 

The `RandomWaveformInjection` object now only injects waveforms into background with some probability. I'm curious to hear thoughts on if we should keep this in `ml4gw`, especially considering that the `BBHnet` version will now override much of the core functionality. 
Personally, I think it makes the most sense to remove it and make it into a prototypical example (via a notebook or something) of how to compose these modules. We could begin by composing the `WaveformSampler` and `WaveformProjector`, continue to add complexity via the `SnrRescaler`, the ability to return sampled parameters (i.e. what the PE project does), some of the waveform augmentations etc... 

Separately, curious to hear feedback on the organizational structure of certain objects (i.e. where should they live).

alecgunny:
@EthanMarx Haven't had a chance to go through this in a ton of detail but looks really good. I think I agree that removing it makes the most sense, and a prototype example is the best way of showing people how to string these objects together to do whatever task it is they want to do, possibly implementing in their own util libraries so they're not constantly reusing code.

For adding the `device` arg to the distribution modules, I don't know that this API is cleaner than just asking that downstream applications know what devices they need things on and call `.to(device)` on whatever the output of the distribution is. Certainly sampling on the GPU is (slightly) faster, but you can handle this more generally by making `"cuda"` the default torch tensor type (maybe including this in a canonical example as well will be helpful for folks).

EthanMarx:
The "core" functionality is all in here and tested (for the most part). 

Before we merge, probably would be a good exercise to recreate the entire`BBHnetRandomWaveformInjection` (and maybe even PE version) with these objects to diagnose any unforeseen failure modes / see where there could be improvements.

ml4gw/transforms/snr_rescaler.py
```diff
@@ -12,16 +12,19 @@ class SnrRescaler(FittableTransform):
     def __init__(
         self,
         n_ifos: int,
```
alecgunny, line 14: I don't know if this comes up anywhere else in the repo, but JIC it does we should land on a standardization between using `n_` and `num_`

ml4gw/transforms/snr_rescaler.py
```diff
@@ -12,16 +12,19 @@ class SnrRescaler(FittableTransform):
...
-        waveform_size: int,
-        highpass: float,
         sample_rate: float,
+        waveform_duration: float,
+        highpass: Optional[float] = None,
         distribution: Optional[Callable] = None,
     ):
+        super().__init__()
         self.distribution = distribution
```
alecgunny, line 21: Would it make sense to include `target_snrs` as an argument of the `forward` method and relegate the sampling to a different `Module`?

alecgunny: And I guess make it `Optional` so that obviously if it's `None` we do the SNR shuffling

ml4gw/transforms/snr_rescaler.py
EthanMarx: I can take a look

ml4gw/transforms/snr_rescaler.py
EthanMarx: Agreed, cleaner and more moduler.

wbenoit26 (approved):

PR #55, opened 2023-07-06
Highpass device fix

alecgunny:
When passing `highpass` as a float to the various SNR-related functions, a device issue gets raised if the responses are on GPU. This fixes that error and switches the `background` argument to `psd` to emphasize what it's meant to represent (as opposed to a time domain representation or even the ASD).


alecgunny:
The other reason I've highlighted the `psd` kwarg here, as I'll note in a couple other PRs, is that since we generally only end up caring about the ASD, which also has the benefit of being much more numerically stable, it might be worth thinking about making this our default frequency domain representation.

wbenoit26 (approved):

PR #57, opened 2023-07-06
Local whitening

alecgunny:
Generalizing the local whitening functionality from Aframe. Todos:
- [x] Figure out normalization
- [x] Work out testing best practices and get tests passing
- [x] Add `Module` wrapper to function that looks like the one in Aframe and just calls these lower level functions

### DIFF
--- a/ml4gw/spectral.py
+++ b/ml4gw/spectral.py
@@ -15,6 +15,9 @@ import numpy as np
 import torch
 from gwpy.frequencyseries import FrequencySeries
 from gwpy.timeseries import TimeSeries
+from torchtyping import TensorType
+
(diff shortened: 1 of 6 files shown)


EthanMarx:
gwpys [whitenening tests](https://github.com/gwpy/gwpy/blob/main/gwpy/timeseries/tests/test_timeseries.py#L1285) are pretty basic: 

Beyond testing for unit variance and mean 0 when whitening gaussian noise, it tests that when a loud signal/glitch is present, the time of the max value of the whitened data corresponds to the time of the max value of the unwhitened data.

EthanMarx (approved):

PR #59, opened 2023-07-11
Chunked dataloader

alecgunny:
Adding a dataloader for loading chunks of segments from multiple files asynchronously
- [ ] ~~Add `shuffle: bool = False` kwarg~~
- [x] Add unit tests
- [x] Add import to `dataloading.__init__`

### DIFF
--- a/ml4gw/dataloading/chunked_dataset.py
+++ b/ml4gw/dataloading/chunked_dataset.py
@@ -158,16 +158,31 @@ class ChunkedDataset(torch.utils.data.IterableDataset):
         coincident: bool = True,
         num_workers: int = 0,
         device: str = "cpu",
+        pin_memory: bool = False,
     ) -> None:
         if not num_workers:
             reads_per_worker = reads_per_chunk
+        elif reads_per_chunk < num_workers:
+            raise ValueError(
+                "Too many workers {} for number of reads_per_chunk {}".format(
+                    num_workers, reads_per_chunk
+                )
+            )
         else:
             reads_per_worker = int(reads_per_chunk // num_workers)
+
+        if kernel_length > chunk_length:
+            raise ValueError(
+                "Kernel length {} must be shorter than "
+                "chunk length {}".format(kernel_length, chunk_length)
+            )
+        self.kernel_size = int(kernel_length * sample_rate)
+        self.chunk_size = int(chunk_length * sample_rate)
 
         chunk_loader = ChunkLoader(
             fnames,
             channels,
-            int(chunk_length * sample_rate),
+            self.chunk_size,
             reads_per_worker,
             chunks_per_epoch,
             coincident=coincident,
@@ -180,7 +195,7 @@ class ChunkedDataset(torch.utils.data.IterableDataset):
                 chunk_loader,
                 batch_size=num_workers,
                 num_workers=num_workers,
-                pin_memory=True,
+                pin_memory=pin_memory,
                 collate_fn=chunk_loader.collate,
             )
 
@@ -188,8 +203,6 @@ class ChunkedDataset(torch.utils.data.IterableDataset):
         self.num_channels = len(channels)
         self.coincident = coincident
 
-        self.chunk_size = int(chunk_length * sample_rate)
-        self.kernel_size = int(kernel_length * sample_rate)
         self.batch_size = batch_size
         self.batches_per_chunk = batches_per_chunk
         self.chunks_per_epoch = chunks_per_epoch
--- a/ml4gw/dataloading/chunked_dataset.py
+++ b/ml4gw/dataloading/chunked_dataset.py
@@ -1,0 +1,158 @@
+from dataclasses import dataclass
+from typing import List
+
+import h5py
+import numpy as np
+import torch
+
+
+@dataclass
+class ChunkLoader:
+    fnames: List[str]
+    channels: List[str]
+    chunk_size: int
+    reads_per_chunk: int
+    chunks_per_epoch: int
+    coincident: bool = True
+
+    def __post_init__(self):
+        sizes = []
+        for f in self.fnames:
+            with h5py.File(f, "r") as f:
+                size = len(f[self.channels[0]][:])
+                sizes.append(size)
+        total = sum(sizes)
+        self.probs = np.array([i / total for i in sizes])
+
+    def __len__(self):
+        return self.chunks_per_epoch
+
+    def sample_fnames(self):
+        return np.random.choice(
+            self.fnames,
+            probs=self.probs,
+            size=(self.reads_per_chunk,),
+            replace=True,
+        )
+
+    def load_coincident(self):
+        fnames = self.sample_fnames()
+        chunks = []
+        for fname in fnames:
+            with h5py.File(fname, "r") as f:
+                chunk, idx = [], None
+                for channel in self.channels:
+                    if idx is None:
+                        end = len(f[channel]) - self.chunk_size
+                        idx = np.random.randint(0, end)
+                    x = f[channel][idx : idx + self.chunk_size]
+                    chunk.append(x)
+            chunks.append(np.stack(chunk))
+        return np.stack(chunks)
+
+    def load_noncoincident(self):
+        chunks = []
+        for channel in self.channels:
+            fnames = self.sample_fnames()
+            chunk = []
+            for fname in fnames:
+                with h5py.File(fname, "r") as f:
+                    end = len(f[channel]) - self.chunk_size
+                    idx = np.random.randint(0, end)
+                    x = f[channel][idx : idx + self.chunk_size]
+                    chunk.append(x)
+            chunks.append(np.stack(chunk))
+        return np.stack(chunks, axis=1)
+
+    def iter_epoch(self):
+        for _ in range(self.chunks_per_epoch):
+            if self.coincident:
+                yield self.load_coincident()
+            else:
+                yield self.load_noncoincident()
+
+    def __iter__(self):
+        return self.iter_epoch()
+
+
+class ChunkedDataloader(torch.utils.data.IterableDataset):
+    """
+    Iterable for generating batches of background data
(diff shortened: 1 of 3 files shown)


ml4gw/dataloading/chunked_dataset.py
```diff
@@ -0,0 +1,267 @@
...
+    def load_coincident(self):
+        fnames = self.sample_fnames()
+        chunks = []
+        for fname in fnames:
+            with h5py.File(fname, "r") as f:
+                chunk, idx = [], None
+                for channel in self.channels:
+                    if idx is None:
+                        end = len(f[channel]) - self.chunk_size
+                        idx = np.random.randint(0, end)
```
wbenoit26, line 50: Could be cleaner to have a single load function and do something like
```
for channel in self.channels:
    end = len(f[channel]) - self.chunk_size
    if idx is None or not coincident: 
        idx = np.random.randint(0, end)
    x = f[channel][idx : idx + self.chunk_size]

```

alecgunny: Yeah been thinking of how to combine these but the issue is that for non-coincident sampling, you kind of flip the order: instead of sampling a bunch of files up front and then reading each channel from them, you loop through the _channels_ then sample a different set of files for each one.

wbenoit26: Ah, yeah, I see, I hadn't noticed the change in order between the two functions. What about something like this?

```
def load(self):
        chunks = []
        for channel in self.channels:
            if channel == self.channels[0] or not coincident:
                 fnames = self.sample_fnames()
            chunk, idx = [], None
            for fname in fnames:
                with h5py.File(fname, "r") as f:
                    end = len(f[channel]) - self.chunk_size
                    if idx is None or not coincident:
                        idx = np.random.randint(0, end)
                    x = f[channel][idx : idx + self.chunk_size]
                    chunk.append(x)
            chunks.append(np.stack(chunk))
        return np.stack(chunks, axis=1)
```

alecgunny: Ok yeah I like this. The one issue is linters will complain that `fnames` is only defined under an `if` statement, and they have no way of knowing that the `if` will _always_ be triggered and so will warn of potential `NameError`s, but that's an easy enough fix (that's why I do the dummy `idx = None`). I'll get this implemented ASAP

alecgunny: FWIW, tried to implement this but realized there's an issue with the `idx`: you need to keep the same indices for each channel when you sample coincidently, but this implementation samples new indices for each channel within a given file. Played around with making this work via `dict`s, but ultimately became more onerous and less readable than just implementing as two separate functions, but happy to take a look if you have a clean implementation.

alecgunny:
Outside of the issue discussed above, this should be good to go. Not the most exhaustive tests in the world, but they get the basic functionality in there. Note that I opted _not_ to implement a non-shuffled functionality, since I don't think it quite makes sense in this setting (where are the chunks from? Contiguous chunks from the same file? At that point would you just load a larger chunk from that file? Or chunks loaded in order from contiguous files? I don't really know what the use case is for that, but if someone thinks of one down the line we can think about how that might fit in here)

ml4gw/dataloading/chunked_dataset.py
wbenoit26: How about defining `self.sizes = sizes` up in `__init__`? And then here you could do

```
chunks, fnames = [], None
for channel in self.channels:
    if fnames is None or not self.coincident:
        fnames = self.sample_fnames()
        sizes = self.sizes[np.argwhere(self.fnames == fnames)]
        ends = [size - self.chunk_size for size in sizes]
        idxs = [np.random.randint(0, end) for end in ends]
    chunk = []
    for idx, fname in zip(idxs, fnames):
        with h5py.File(fname, "r") as f:
            x = f[channel][idx : idx + self.chunk_size]
            chunk.append(x)
        chunks.append(np.stack(chunk))
return np.stack(chunks, axis=1)
```

tests/dataloading/test_chunked_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+        if num_workers:
+            assert dataset.chunk_loader.dataset.reads_per_chunk == 1
+
+    def test_iter(self, dataset, coincident):
+        for i, x in enumerate(dataset):
+            assert x.shape == (8, 2, 192)
+            if coincident:
+                torch.testing.assert_close(x[:, 0], -x[:, 1], rtol=0, atol=0)
+            else:
+                assert (x[:, 0] != -x[:, 1]).any().item()
```
wbenoit26, line 191: I think this should be `.all()` like it is above.

ml4gw/dataloading/chunked_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+                time_idx = time_idx.view(self.batch_size, -1, 1)
+
+                # there's no additional offset factor to account for here
+                chunk_idx += time_idx
+
+                # now slice this 3D tensor from our flattened chunk
+                yield chunk[chunk_idx]
+
+    def __iter__(self):
+        return self.iter_epoch()
```
wbenoit26, line 280: Question for my own sake: why make a separate function if all `__iter__` does is call it?

tests/dataloading/test_chunked_dataset.py
alecgunny: No because it's possible (though unlikely) that a segment gets paired with its coincident counterpart (though maybe it's so unlikely that it's worth failing this test one out of however many times to have a more thorough check)

ml4gw/dataloading/chunked_dataset.py
alecgunny: This is a great question. The one real requirement of `__iter__` is that it return an object with a `__next__` method. That's what's functionally happening under the hood when you do e.g. `for i in iterable`: Python calls the `__iter__` method on `iterable`, then at each step in the loop assigns to `i` the output of the `__next__` method of the object that gets returned by `__iter__` (the `iterator`, not necessarily `iterable` _itself_). This repeats until calling `__next__` raises a `StopIteration`.

So an `Iterable` is an object with an `__iter__` method that shouldn't do the iteration itself, but rather return an `Iterator`, an object with a `__next__` method. Some objects will act as both `Iterable` and `Iterator`, using `__iter__` to initialize some iteration parameters and then `return self`, so that its own `__next__` method is what gets called at each loop. Each call to `__next__` will typically increment those parameters. Take for example the following dummy implementation of `range`

```python
class Range:
    def __init__(self, start, stop, step):
        self.start = start
        self.stop = stop
        self.step = step
        self._i = None

    def __iter__(self):
        self._i = self.start
        return self

    def __next__(self):
        if self._i >= self.stop:
            self._i = None
            raise StopIteration
        value = self._i + 0
        self._i = self._i + self.step
        return value
```
This is fine, but as is noted in [this](https://dbader.org/blog/python-iterators) great overview: "Python iterators normally can’t be 'reset'—once they’re exhausted they’re supposed to raise `StopIteration` every time `next()` is called on them." So a common pattern is to break up the `Iterable` and the `Iterator` into separate objects, such that every time you call `iter` on the `Iterable`, it returns a fresh `Iterator` that you can loop through.

So all I'm doing here is rather than

- creating a dedicated `Iterator` object that would have to borrow a bunch of attributes from this object
- creating a `__next__` method where I have to keep track of incrementing indices and raising `StopIteration` myself, and technically violating this Python rule-of-thumb (since we'll iterate through this dataset several times)

I create a generator using an internal method, and return that as my `Iterator`.

tests/dataloading/test_chunked_dataset.py
wbenoit26: Yeah, I think better to have the more thorough check. At least, we should be consistent between this and the check on line 99.

ml4gw/dataloading/chunked_dataset.py
wbenoit26: Got it, okay, thanks for explaining all that.

tests/dataloading/test_chunked_dataset.py
alecgunny: Yeah I think the reason I had those different was that the odds of having the chunks match up are way lower, since you're taking a smaller fraction of the full files (1/5, 1/2, and 1/3, respectively). But the odds of having two _kernels_ match up is much higher since the kernels occupy such a large fraction of the chunks. But agreed they should be consistent.

I would, however, make the argument that the `.any()` check is the correct one: once you've proved that when `coincident is True` that everything matches up, the only thing you can really require about when `coincident is False` is that at least _something_ doesn't match up. Then you know you're getting the behavior your expect. If some of them do match up, well that's still expected behavior some of the time. (So is all of them matching up I guess, but those odds are so low that I'm not gonna worry about it)

wbenoit26: Hmm, yeah, I'll buy that.

ml4gw/dataloading/chunked_dataset.py
wbenoit26: Did you get a chance to take a look at this?

tests/dataloading/test_chunked_dataset.py
alecgunny: Ok cool, this should be consistent now then

ml4gw/dataloading/chunked_dataset.py
alecgunny: For the record: Will and I chatted over Slack about the downside of this approach being that you're doing unnecessary extra file openings inside the loop for the `coincident=True` case, which can introduce non-trivial latency, especially for shorter chunks. Probably easier to keep these separate for now, but also raises the question of whether we should introduce a parameter for the user to control whether non-coincident sampling reads from independent files or from independently sampled times from the _same_ file to avoid this latency as well. Will file an issue for this with the relevant latency data for reference.

wbenoit26 (approved):

PR #60, opened 2023-07-11
Python bump

alecgunny:
Placeholder PR for eventually bumping python support and retiring gwpy dependency. Gwpy dependency should be officially gone when #57 is merged and we no longer need `normalize_background`, which I've removed here even though it still gets called in the `Whitening` module.

### DIFF
--- a/ml4gw/spectral.py
+++ b/ml4gw/spectral.py
@@ -9,12 +9,9 @@ and
 https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.csd.html
 """
 
-from typing import Optional, Union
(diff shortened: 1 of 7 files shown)


alecgunny:
@EthanMarx scope of `SnrRescaler` tests has been massively reduced since we've discussed plan to deprecate this anyway, and all the existing tests do is essentially make sure that `gw.compute_network_snr` is being called correctly. I can get the relevant objects implemented before 0.2 release and the tests should end up looking a lot like the whitening tests, so should be pretty easy.

EthanMarx (approved):

PR #62, opened 2023-07-27
Remove leftover `scipy` and `numpy` utils

EthanMarx:
1. Removes `numpy` import and type annotation in `scaler.py`
2. Removes scipy `_median_bias` call in spectral.py

### DIFF
--- a/ml4gw/spectral.py
+++ b/ml4gw/spectral.py
@@ -16,11 +16,6 @@ from torchtyping import TensorType
 
 from ml4gw import types
 
-try:
-    from scipy.signal.spectral import _median_bias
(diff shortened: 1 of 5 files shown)


EthanMarx:
@alecgunny does it make sense to remove support for taking `ndarrays` as arguments? We currently do this in the scalers/whiteners `.fit` methods, and a couple other places. Otherwise, I think we need to add numpy back as a dependency

EthanMarx:
This is ready to go, although a whitening test is failling. I imagine this might get resolved via #63

wbenoit26 (approved):

PR #63, opened 2023-07-27
Looser bounds

alecgunny:
Introducing some logic to set error bounds on failure-prone tests.

Based on #61 , so needs that to get merged then will rebase on top of `dev` from there.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ def validate_whitened():
         target = torch.zeros_like(means)
         torch.testing.assert_close(means, target, rtol=0, atol=0.02)
 
-        stds = whitened.var(axis=-1)
+        stds = whitened.std(axis=-1)
         target = torch.ones_like(stds)
 
         # if we're highpassing, then we shouldn't expect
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -10,18 +10,30 @@ def validate(whitened, highpass, sample_rate, df):
         target = torch.zeros_like(means)
         torch.testing.assert_close(means, target, rtol=0, atol=0.02)
 
-        stds = whitened.std(axis=-1)
+        stds = whitened.var(axis=-1)
```
wbenoit26, line 13: Why did this change if we're comparing to standard deviation below?

alecgunny: The idea here is that your estimate of the _variance_ of an N(0, 1) variable from `N` observations should follow a chi-squared distribution with `N` degrees of freedom (normalized by `1 / N`). The standard deviation of this distribution (the distribution of the estimate of the standard deviation) should be `sqrt(2 / N)`, so the std we're comparing against below is to make sure that all our variance estimates fall within the 5-sigma (or 25, which I set below just to see if I could get things passing consistently) of their expected distribution.

alecgunny: But as we discussed yesterday, there's a bias in our standard deviation estimates being caused by some difference in scaling between what comes out of the `spectral_density` function and the FFT, so things aren't following this distribution correctly at the moment.

wbenoit26 (approved):

PR #65, opened 2023-08-03
Remove numpy in gw and dataloading modules

EthanMarx:
Remove numpy use in `in_memory_dataset`, `gw`, and `whitening` modules. After this we are now fully free of `numpy` dependency


EthanMarx:
@alecgunny Quick PR that gets rid of all uses of `numpy`

wbenoit26 (approved):

PR #67, opened 2023-09-05
reformulating things to build batch-level dataloader

alecgunny:


ml4gw/dataloading/hdf5_dataset.py
```diff
@@ -0,0 +1,175 @@
...
+from typing import Sequence, Union
+
+import h5py
+import numpy as np
+import torch
+
+from ml4gw.types import WaveformTensor
+
+
+class ContiguousHdf5Warning(Warning):
```
EthanMarx, line 11: Is this meant to be passed to `warnings.warn` ?

ml4gw/dataloading/hdf5_dataset.py
```diff
@@ -0,0 +1,175 @@
...
+        fnames: Sequence[str],
+        channels: Sequence[str],
+        kernel_size: int,
+        batch_size: int,
+        batches_per_epoch: int,
+        coincident: Union[bool, str],
+    ) -> None:
+        if not isinstance(coincident, bool) and coincident != "files":
+            raise ValueError(
+                "coincident must be either a boolean or 'file', "
```
EthanMarx, line 66: error message should say "files" not "file"

ml4gw/dataloading/hdf5_dataset.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __iter__(self) -> torch.Tensor:
+        worker_info = torch.utils.data.get_worker_info()
+        if worker_info is None:
+            num_batches = self.batches_per_epoch
+        else:
+            num_batches, remainder = divmod(
+                self.batches_per_epoch, worker_info.num_workers
+            )
+            if worker_info.id < remainder:
```
EthanMarx, line 171: Can you explain the logic here?

ml4gw/dataloading/hdf5_dataset.py
alecgunny: No it's just creating a unique warning type so that downstream applications can catch it explicitly. But I wasn't using it in the warning below so good catch there, added it in now.

ml4gw/dataloading/hdf5_dataset.py
alecgunny: Yup good catch

EthanMarx (approved):

PR #78, opened 2023-11-09
shifted pearson correlation

alecgunny:
Adding an implementation of the Pearson correlation we've had floating around for a while now. Tests are pretty basic but it works for now. Nice to haves, but maybe not strictly required for getting this merged, are:
- [ ] Needs testing for when the first input is 2 or 1 dimensional
- [ ] See TODOs in class doc string: possible support for strides and windowing

### DIFF
--- a/ml4gw/transforms/__init__.py
+++ b/ml4gw/transforms/__init__.py
@@ -1,3 +1,4 @@
+from .pearson import ShiftedPearsonCorrelation
 from .scaler import ChannelWiseScaler
 from .snr_rescaler import SnrRescaler
 from .spectral import SpectralDensity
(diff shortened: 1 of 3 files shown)


wbenoit26 (approved):
I agree that neither of those actions are necessary to get this merged in. They could be good first issues for somebody, though. Why would windowing be desirable here?

alecgunny:
@wbenoit26 I know this is something other libs like gwpy do when computing convolution-based correlations. My grasp of the specifics it tenuous, but basically the signal processing mechanism behind building your zero-padded window is that that window is part of some longer timeseries which you're effectively multiplying by a tophat element-wise. This equates to convolving with a sinc function in the frequency domain, which smears your frequency content and leads to inaccurate results. Applying a window with more desirable spectral properties than the tophat reduces these artifacts.

PR #79, opened 2023-11-09
Initial chi-squared veto implementation

alecgunny:
Generalizes some of the SNR calculation code to make this feasible and plays around with some scaling on this front. Think the transform itself is still missing some normalization constants here and there because the values are coming out enormous, but the general idea seems to work.

This metric should be close to 1 if the observed data contains the signal plus some gaussian noise, so the idea with testing is to construct such a signal/observation combination and test that the values come out as expected, but still not fully resolved how best to do this.

- [ ] Figure out normalization in the transform class
- [ ] Get rid of debugging print statements once things are solved
- [ ] Figure out how best to handle scaling in the SNR calculations, see all the TODOs in `gw.py`

### DIFF
--- a/ml4gw/gw.py
+++ b/ml4gw/gw.py
@@ -285,7 +285,7 @@ def get_ifo_geometry(
     return torch.Tensor(tensors), torch.Tensor(vertices)
 
 
-def snr_from_freqs(
+def snr_frequency_series(
     template: PSDTensor,
     df: float,
     psd: Optional[PSDTensor] = None,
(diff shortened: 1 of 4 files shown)


ml4gw/gw.py
```diff
@@ -285,6 +285,96 @@ def get_ifo_geometry(
...
+    # multiplying by the sqrt of the scale factor.
+    # This is less optimal than our previous implementation
+    # because we're now doing 2 element-wise divides by
+    # the ASD rather than 1, so worth figuring out if
+    # there's a better, more consistent way to make the
+    # scale workout. All see TODO below
+    asd = psd**0.5
+    template = template * scale**0.5 / asd
+    other = other.conj() * scale**0.5 / asd
+    integrand = template * other
```
wbenoit26, line 316: Is this different than doing `integrand = template * other.conj() * scale / psd`? Do you run into precision issues?

alecgunny: yeah that's exactly it, this is the most stable implementation I've found

PR #84, opened 2023-11-30
Autoencoders

alecgunny:
initial library for implementing arbitrary blocks of autoencoders


EthanMarx (approved):
approving and merging to fix dep issue with `aframev2`. Will follow up with an issue to write tests before `dev` get's into main.

PR #85, opened 2023-12-01
TaylorF2 3.5PN in torch

deepchatterjeeligo:


ml4gw/waveforms/taylorf2.py
```diff
@@ -0,0 +1,198 @@
...
+        * PI
+    )
+
+    # Energy coefficients
+    dETaN = 2 * (-eta / 2.0)
+    dETa1 = 2.0 * -(0.75 + eta / 12.0)  # noqa  F841
+    dETa2 = 3.0 * -(  # noqa  F841
+        27.0 / 8.0 - 19.0 / 8.0 * eta + 1.0 / 24.0 * eta * eta
+    )  # noqa  F841
+    dETa3 = 4.0 * -(  # noqa  F841
```
wbenoit26, line 139: Why do all the `FTa#`s and `dETa#`s not get used anywhere?

deepchatterjeeligo: That's because taylorf2 is newtonian order in amplitude. I'll remove them now, we can bring them back when we get phenomD in.

ml4gw/waveforms/taylorf2.py
```diff
@@ -0,0 +1,165 @@
...
+    TaylorF2 up to 3.5 PN in phase. SPA amplitude.
+    params = [mass1, mass2, D, phic, inclination]
+
+    Returns:
+    --------
+      hp, hc
+    """
+    # shape assumed (n_batch, params)
+    # frequency array is repeated along batch
+    inclination = params[:, 4]
```
wbenoit26, line 156: I think it would be cleaner to have `params` be a dictionary with the parameter names as keys

ml4gw/waveforms/taylorf2.py
```diff
@@ -0,0 +1,165 @@
...
+        + eta * eta * 76.055 / 1.728
+        - eta * eta * eta * 127.825 / 1.296
+        + pfa_v6logv * torch.log(torch.tensor(4.0))
+    )
+    pfa_v6logv *= pfaN
+    pfa_v6 *= pfaN
+    pfa_v7 = PI * (
+        770.96675 / 2.54016 + 378.515 / 1.512 * eta - 740.45 / 7.56 * eta * eta
+    )
+    pfa_v7 *= pfaN
```
wbenoit26, line 84: Can we do this multiplication by `pfaN` all at once after `phasing has been calculated?

ml4gw/waveforms/taylorf2.py
```diff
@@ -0,0 +1,165 @@
...
+    phasing = (v7.T * pfa_v7).T
+    phasing += (v6.T * pfa_v6 + v6_logv.T * pfa_v6logv).T
+    phasing += (v5.T * pfa_v5 + v5_logv.T * pfa_v5logv).T
+    phasing += (v4.T * pfa_v4).T
+    phasing += (v3.T * pfa_v3).T
+    phasing += (v2.T * pfa_v2).T
+    phasing += (v1.T * pfa_v1).T
+    phasing += (v0.T * pfa_v0).T
+    # Divide by 0PN v-dependence
+    phasing /= v5
```
wbenoit26, line 95: Just a question for my own benefit: at what power of v do things become post-Newtonian?

ml4gw/waveforms/taylorf2.py
deepchatterjeeligo: Yes. Now I recall I did this to debug the coefficients.

ml4gw/waveforms/taylorf2.py
deepchatterjeeligo: What can be done is to pass explicit variables mass1, mass2, etc. But each should have the shape (n_batch, 1). But then there has to be extra assert statements to check that dim 0 matches for all arguments. I can do that, it depends what is better use case.

ml4gw/waveforms/taylorf2.py
deepchatterjeeligo: The newtonian term in phasing is just 3/128 * eta * v^-5. This is called 0PN or Newtonian. So think of the expansion as [Newtonian] * [1 + (0.5 PN) * v + (1 PN) * v^2 + ...].

ml4gw/waveforms/taylorf2.py
```diff
@@ -0,0 +1,165 @@
...
+    Mf = (f.T * M_s).T
+
+    v0 = torch.ones_like(Mf)
+    v1 = (PI * Mf) ** (1.0 / 3.0)
+    v2 = v1 * v1
+    v3 = v2 * v1
+    v4 = v3 * v1
+    v5 = v4 * v1
+    v6 = v5 * v1
+    v7 = v6 * v1
```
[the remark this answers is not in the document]
wbenoit26: Got it, cool, that makes sense to me

wbenoit26 (approved):

PR #88, opened 2023-12-14
add spin contributions to PN coeffs

deepchatterjeeligo:


ml4gw/waveforms/taylorf2.py
```diff
@@ -69,9 +122,69 @@ def taylorf2_phase(
         - eta * eta * eta * 127.825 / 1.296
         + pfa_v6logv * torch.log(torch.tensor(4.0))
     )
+    # SO + S1-S2 + S-squared contribution at 3 PM
```
wbenoit26, line 125: Minor comment that I won't hold this up for: this comment should say `3PN` rather than `3PM`

wbenoit26 (approved):

ml4gw/waveforms/taylorf2.py
```diff
@@ -52,13 +58,60 @@ def taylorf2_phase(
...
+        + (240.0 / 9.6 * m2byM * m2byM + -7.0 / 9.6 * m2byM * m2byM) * chi2sq
+    )
     pfa_v5logv = 5.0 / 3.0 * (772.9 / 8.4 - 13.0 * eta) * PI
     pfa_v5 = 5.0 / 9.0 * (772.9 / 8.4 - 13.0 * eta) * PI
+    # SO coefficient for 2.5 PN
+    pfa_v5logv += 3.0 * (
+        -m1byM
+        * (
+            1391.5 / 8.4
+            - m1byM * (1.0 - m1byM) * 10.0 / 3.0
```
alecgunny, line 87: Sorry this is too exciting of a development and I had to take a quick peek. My only two cents I'll throw in is that, depending on where your bottlenecks are, it could be worth being deliberate about how you include constants here. For example, this will first compute `m1byM * (1.0 - m1byM)`, which will be a tensor, then multiply that element-wise by 10, then divide that tensor element-wise by 3. So you're paying for an extra kernel launch and element-wise multiply that you wouldn't need if you did `10.0 / 3.0. * m1byM * (1.0 - m1byM)`. A small thing, but could add up in the aggregate, especially for some of the square terms you could pre-compute. Anyway, minor point, this is extremely cool.

wbenoit26: Oh interesting, yeah, we should do a timing comparison and see what sort of speed up we can get. Could be worth deviating from lalsuite for.

deepchatterjeeligo: OK I did some quick timing test locally. With lalsimulation, getting waveforms in sequence takes around 3 milliseconds per waveform generation. With the torch implementation, a batch of 1000 waveforms is created in 500 ms i.e. 0.5 ms per waveform. This is on my laptop that does not have a GPU.

```
%%timeit -n 100
hp1, hc1 = lalsimulation.SimInspiralChooseFDWaveform(**pars)

2.9 ms ± 275 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

vs.

```
params = torch.tensor([m1, m2, chi1, chi2, D, phic, inclination])

params = params.repeat([1000, 1])
params.shape # torch.Size([1000, 7])

%%time
hp_torch, hc_torch = taylorf2.TaylorF2(
    torch_freqs,
    params[:,0],
    params[:,1],
    params[:,2],
    params[:,3],
    params[:,4],
    params[:,5],
    params[:,6],
    f_ref
)

CPU times: user 4.55 s, sys: 2.12 s, total: 6.67 s
Wall time: 511 ms
```

deepchatterjeeligo: I moved the scalar pre-factors ahead.

wbenoit26: Nice! I tried computing all the repeated tensor products up front, and saw a small improvement there as well. However, it involved a lot more rearranging, and I think it's not worth deviating much from lalsim at this point.

alecgunny: Interesting, yeah 10% seems about appropriate, though you might observe a bit more acceleration on GPU where the compute time is fast enough to be comparable to the kernel launch overhead. But either way, insanely stoked to see what you guys do with this.

PR #89, opened 2023-12-22
add batched IMRPhenomD

deepchatterjeeligo:


wbenoit26:
It looks like this is missing the `phenom_d_data` file

tests/waveforms/test_cbc_waveforms.py
```diff
@@ -131,3 +141,96 @@ def test_taylor_f2(
...
+    torch_mask = (torch_freqs > params["f_min"]) & (
+        torch_freqs < params["f_max"]
+    )
+
+    hp_lal_data = hp_lal.data.data[lal_mask]
+    hc_lal_data = hc_lal.data.data[lal_mask]
+    hp_torch = hp_torch[torch_mask]
+    hc_torch = hc_torch[torch_mask]
+
+    assert np.allclose(
```
wbenoit26, line 225: I'm guessing that the tolerance here is to account for the precision issues you were seeing around the amplitude?

deepchatterjeeligo: you're right.

wbenoit26 (approved):

PR #90, opened 2024-01-23
Added multi-resolution spectrogram module

wbenoit26:
Adds a transform that computes a single spectrogram by calculating multiple spectrograms at different resolutions and taking the maximum each corresponding pixel.

- [X] Add shape checks
- [X] Add tests
- [X] Finish writing docstring
- [X] Allow specification of more parameters than just `n_fft`

### DIFF
--- a/ml4gw/transforms/spectrogram.py
+++ b/ml4gw/transforms/spectrogram.py
@@ -58,6 +58,15 @@ class MultiResolutionSpectrogram(torch.nn.Module):
         self.num_freqs = max([shape[0] for shape in self.shapes])
         self.num_times = max([shape[1] for shape in self.shapes])
 
+        bottom_pad = torch.tensor(
+            [int(self.num_freqs - shape[0]) for shape in self.shapes]
+        )
+        right_pad = torch.tensor(
+            [int(self.num_times - shape[1]) for shape in self.shapes]
+        )
+        self.register_buffer("bottom_pad", bottom_pad)
+        self.register_buffer("right_pad", right_pad)
+
         freq_idxs = torch.tensor(
             [
                 [int(i * shape[0] / self.num_freqs) for shape in self.shapes]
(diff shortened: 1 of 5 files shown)


wbenoit26:
@EthanMarx Are you aware of any issues that might come from bumping our `torch` version? Looks like `torchaudio` 0.13 is incompatible with python 3.11, but the next version of `torchaudio` requires `torch > 2.0`

ml4gw/transforms/spectrogram.py
```diff
@@ -0,0 +1,138 @@
...
+        if X.shape[-1] != self.kernel_size:
+            raise ValueError(
+                "Expected time dimension to be "
+                f"{self.kernel_size} samples long, got input with "
+                f"{X.shape[-1]} samples"
+            )
+
+        spectrograms = [t(X) for t in self.transforms]
+
+        left_pad = torch.zeros(len(spectrograms), dtype=torch.int)
```
EthanMarx, line 126: Can these be made buffers in the`__init__` as well?

tests/transforms/test_spectrogram.py
```diff
@@ -0,0 +1,111 @@
...
+
+    X = torch.ones([batch_size, num_channels, kernel_length * sample_rate])
+    y = spectrogram(X)
+    expected_y = Spectrogram(normalized=True)(X)
+
+    assert (y == expected_y).all()
+
+    # The `normalized = False` should be ignored
+    spectrogram = MultiResolutionSpectrogram(
+        kernel_length, sample_rate, normalized=[False]
```
EthanMarx, line 80: I think it would be better to raise a `ValueError` in this instance to let the user know this isn't possible. Otherwise the user will silently get unexpected behavior.

EthanMarx (changes requested):
Look good, might also be worth adding a warning with `warnings.warn` if the total number of spectrograms is larger than some value since we see some pretty poor computational scaling.

tests/transforms/test_spectrogram.py
wbenoit26: Agreed.

EthanMarx:
> @EthanMarx Are you aware of any issues that might come from bumping our `torch` version? Looks like `torchaudio` 0.13 is incompatible with python 3.11, but the next version of `torchaudio` requires `torch > 2.0`

Assuming all of our tests pass I don't think so. If anything, it would affect downstream projects. I guess this is a good reason to keep it in `dev` for a bit until we see potential issues.

EthanMarx (approved):

PR #92, opened 2024-01-25
Augmentations

EthanMarx:
Adds `SignalInverter` and `SignalReverser` augmentations

### DIFF
--- a/tests/test_augmentations.py
+++ b/tests/test_augmentations.py
@@ -6,11 +6,6 @@ import torch
 from ml4gw.augmentations import SignalInverter, SignalReverser
 
 
-@pytest.fixture(params=[0, 0.25, 1])
-def prob(request):
-    return request.param
-
-
 @pytest.fixture(params=[0.0, 0.25, 0.5, 1])
 def flip_prob(request):
     return request.param
--- a/tests/test_augmentations.py
+++ b/tests/test_augmentations.py
@@ -1,0 +1,118 @@
+from unittest.mock import patch
+
+import pytest
+import torch
+
+from ml4gw.augmentations import SignalInverter, SignalReverser
+
+
+@pytest.fixture(params=[0, 0.25, 1])
+def prob(request):
+    return request.param
+
+
+@pytest.fixture(params=[0.0, 0.25, 0.5, 1])
+def flip_prob(request):
+    return request.param
+
+
+@pytest.fixture
+def rvs():
+    return torch.Tensor([[0.0, 0.49], [0.51, 0.1], [0.9, 0.2], [0.1, 0.3]])
+
+
+@pytest.fixture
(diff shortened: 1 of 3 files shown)


tests/test_augmentations.py
```diff
@@ -0,0 +1,118 @@
+from unittest.mock import patch
+
+import pytest
+import torch
+
+from ml4gw.augmentations import SignalInverter, SignalReverser
+
+
+@pytest.fixture(params=[0, 0.25, 1])
+def prob(request):
```
wbenoit26, line 10: Does this get used anywhere?

tests/test_augmentations.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if pos_idx:
+        pos0, pos1 = zip(*pos_idx)
+        assert (X[pos0, pos1] == false).all()
+    elif prob != 1:
+        raise ValueError("Missing positive indices")
+
+
+def test_signal_inverter(flip_prob, rvs, true_idx):
+    tform = SignalInverter(prob=0.0)
```
wbenoit26, line 76: Do you still need `prob = 0` and `prob = 1` as separate cases for both of these?

tests/test_augmentations.py
```diff
@@ -0,0 +1,118 @@
...
+    X = tform(X)
+
+    assert (X == -Y).all()
+
+    tform = SignalInverter(flip_prob)
+    X = torch.ones((4, 2, 8))
+    with patch("torch.rand", return_value=rvs):
+        X = tform(X)
+    X = X.cpu().numpy()
+    validate_augmenters(X, true_idx, -1, 1, flip_prob)
```
wbenoit26, line 94: Why are the true/false values -1 and 1 if the values in `X` come from a normal distribution?

EthanMarx: (here they come from torch.ones)

tests/test_augmentations.py
EthanMarx: No, but I explicitly added them since they're more readable / intuitive, and also  as queue for future self to refactor the. more complicated tests

tests/test_augmentations.py
EthanMarx: Yes in validate_augmentors I think

tests/test_augmentations.py
wbenoit26: Ah, I see now

tests/test_augmentations.py
wbenoit26: Does it? It looks like that's just getting passed the `flip_prob` argument from each of the test functions

EthanMarx: Done

wbenoit26 (approved):

PR #97, opened 2024-02-01
White noise

EthanMarx:
1. Adds functionality for generating colored white noise from PSD
2. Changes `waveforms` package to `simulation`. Adds `cbc` submodule where IMRphenom and taylorF2 now live

- [ ] Tests for white noise generator


EthanMarx:
@deepchatterjeeligo Let me know what you think about this minor  `cbc` submodule structure change

wbenoit26:
This needs to be switched to be against `dev`

PR #98, opened 2024-02-01
Resnet 2d

EthanMarx:
Adds `ResNet2D` that is essentially the same as `ResNet1D` but with Conv2D layers. Probably room to generalize these into one object that takes the convolutional layer type as a parameter, but will leave that to future work. 

The ResNet2D uses the `pytorch` `GroupNorm` implementation which is known to be slower at inference time. Generalizing our faster GroupNorm1D  is also left as an issue.


wbenoit26 (changes requested):
Other than the notebook, this looks good to me

wbenoit26 (approved):

PR #110, opened 2024-02-21
loosely re-implement distributions using torch.distributions

deepchatterjeeligo:
I was thinking of having a loose implementation of the bilby priors using torch tools, to directly sample from priors on the GPU. One intermediary step is to take advantage of the torch distributions and transforms. 

Several of the distributions like `Uniform` and `LogUniform` do not need to be separately implemented in `ml4gw` since they already exist in `torch.distributions`. I did not remove it here yet (since libs downstream may still be using it), but can do so if we agree. The others like PowerLaw, Sine, and Cosine is re-implemented as a Distribution or a Transformed-distribution.

### DIFF
--- a/ml4gw/distributions.py
+++ b/ml4gw/distributions.py
@@ -103,17 +103,22 @@ class LogNormal:
         return x
 
 
-class LogUniform(Uniform):
+class LogUniform(dist.TransformedDistribution):
     """
     Sample from a log uniform distribution
     """
 
-    def __init__(self, low: float, high: float) -> None:
-        super().__init__(math.log(low), math.log(high))
-
-    def __call__(self, N: int) -> torch.Tensor:
-        u = super().__call__(N)
-        return torch.exp(u)
+    def __init__(self, low: float, high: float, validate_args=None):
+        base_dist = dist.Uniform(
+            torch.as_tensor(low).log(),
+            torch.as_tensor(high).log(),
+            validate_args,
+        )
+        super().__init__(
+            base_dist,
+            [dist.ExpTransform()],
+            validate_args=validate_args,
+        )
 
 
 class PowerLaw(dist.TransformedDistribution):
--- a/ml4gw/distributions.py
+++ b/ml4gw/distributions.py
@@ -9,6 +9,7 @@ import math
 from typing import Optional
 
 import torch
+import torch.distributions as dist
 
 
 class Uniform:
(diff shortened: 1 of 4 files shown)


wbenoit26:
Great idea, looks like there's constraints as well that we can use.

ml4gw/distributions.py
```diff
@@ -47,6 +48,59 @@ def __call__(self, N: int) -> torch.Tensor:
         return torch.arcsin(u / self.norm + math.sin(self.low))
 
 
+class CosineDistribution(dist.Distribution):
```
EthanMarx, line 51: I think for sake of naming consistency with the other `torch` distributions ( as well as our own ) we should name this `Cosine` instead of `CosineDistribution`

ml4gw/distributions.py
```diff
@@ -47,6 +48,59 @@ def __call__(self, N: int) -> torch.Tensor:
...
+        u = torch.rand(sample_shape)
+        return torch.arcsin(u / self.norm + math.sin(self.low))
+
+    def log_prob(self, value):
+        value = torch.as_tensor(value)
+        inside_range = (value >= self.low) & (value <= self.high)
+        return value.cos().log() * inside_range
+
+
+class SineDistribution(dist.TransformedDistribution):
```
EthanMarx, line 80: See above

ml4gw/distributions.py
```diff
@@ -120,3 +174,33 @@ def __call__(self, N: int) -> torch.Tensor:
         samples = self.x_min ** (-self.alpha + 1) - u
         samples = torch.pow(samples, -1.0 / (self.alpha - 1))
         return samples
+
+
+class PowerLawDistribution(dist.TransformedDistribution):
```
EthanMarx, line 179: See above

EthanMarx (changes requested):
Thanks for addressing this Deep. I think we can probably remove the other non-torch dists at this point, since most projects point use `ml4gw` as a submodule and point to a specific commit.

EthanMarx:
@deepchatterjeeligo Another question. Do you know what the `validate_args` parameter does? I couldn't find documentation on it

ml4gw/distributions.py
deepchatterjeeligo: Right, so that's something I wanted to ask. Note that the API is slightly different. For example, one has to call `.sample(...)` instead of the class object being callable. So downstream, one will have to do
```
cos = Cosine()
cos_samples = cos.sample((1000,))
```
instead of
```
cos = Cosine()
cos_samples = cos(10000)
```
But its a small enough change to be done in downstream libs.

deepchatterjeeligo:
Practically, I have experienced differing dtypes raising an error with Normal distribution with validate_args set to True. But I think one can re-implement the validation.

ml4gw/distributions.py
deepchatterjeeligo: Done

ml4gw/distributions.py
deepchatterjeeligo: Done

EthanMarx (approved):

PR #112, opened 2024-03-04
Add bilby_prior_equivalent to all distributions used

deepchatterjeeligo:


wbenoit26:
I'm wondering if it would be reasonable to expect downstream projects to subclass these distributions themselves if they wanted this functionality. This seems like kind of a specific use case. @EthanMarx, do you have any thoughts here?

EthanMarx:
I think I would agree that this should probably stay as a downstream usecase, at least for now. In general, I think we should decouple from libs like `bilby` as much as possible.

EthanMarx:
@deepchatterjeeligo gonna close this for now b/c I think we have a nice workaround in amplfi.

PR #113, opened 2024-03-08
Q-transform

wbenoit26:


wbenoit26:
@EthanMarx Can you take a look at this when you get a chance? I want to get this in so that I can use it for the Aframev2 BNS PR

ml4gw/transforms/qtransform.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def compute_qtiles(self, X: torch.Tensor, norm: str = "median"):
+        """
+        Take the FFT of the input timeseries and calculate the transform
+        for each `QTile`
+        """
+        # Computing the FFT with the same normalization and scaling as GWpy
+        X = torch.fft.rfft(X, norm="forward")
+        X[..., 1:] *= 2
+        self.qtiles = [qtile(X, norm) for qtile in self.qtile_transforms]
```
EthanMarx, line 270: Is there a reason for updating this attribute in place as oppposed to returning the qtiles, and letting downstream functions handle how to process it?

Mostly just a readability / code-clarity thing

wbenoit26: It's not updating in place, `self.qtile_transforms` is the list of transformations and `self.qtiles` is the list of transformed data (though, I see what you mean about readability with those two names). I think my thought process had been that the transformed data should live with the Q-transform object once it had been computed, so I wanted attributes for both the transforms and the transformed data.

I could have this function return `self.qtiles` in addition to assigning the attribute.

ml4gw/transforms/qtransform.py
```diff
@@ -0,0 +1,444 @@
...
+                for the maximum energy
+            norm:
+                The method of interpolation used by each QTile
+
+        Returns:
+            An interpolated Q-transform for the batch of data. Output will
+            have one more dimension than the input
+        """
+        for transform in self.q_transforms:
+            transform.compute_qtiles(X, norm)
```
EthanMarx, line 435: Continuing above discussion: here is where my confusion stemmed from. Was unclear what was data was being manipulated / adjusted to do the transform. 

Won't hold this up if you think this is the best way to do it.

EthanMarx (approved):

ml4gw/transforms/qtransform.py
wbenoit26: Okay, cool. I'll merge this in, but keep this in mind for future updates

PR #116, opened 2024-04-22
Generalize Q Transform Input Dimensionality

### LINKED ISSUE #115: Allow Q-transform to accept > 3 dimensional inputs
It should be doable to generalize the current Q-transform to act on tensors of arbitrary dimension. This would be useful, for example, in Aframe where the foreground validation set is 4-dimensional.

wbenoit26:
Thanks to the tip from Charline, the normalization step is now implemented in an intelligent way, and allows the input to be of any dimension.

Edit: I've realized that torch's interpolation doesn't allow for arbitrary dimensions, so the Q transform input can't be of arbitrary dimension either. However, re-writing the normalization be more readable is still good.

### DIFF
--- a/ml4gw/transforms/qtransform.py
+++ b/ml4gw/transforms/qtransform.py
@@ -15,7 +15,7 @@ input on GPU.
 class QTile(torch.nn.Module):
     """
     Compute the row of Q-tiles for a single Q value and a single
-    frequency for multi-dimensional frequency series data.
+    frequency for a batch of multi-channel frequency series data.
     Should really be called `QRow`, but I want to match GWpy.
     Input data should have three dimensions or fewer.
     If fewer, dimensions will be added until the input is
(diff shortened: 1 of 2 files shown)


EthanMarx (approved):
Awesome - to be fair to yourself you did "swear there is a better way to do this". `keepdim` is very helpful when you need to do array broadcasting.

I'm going to assume the interpolation changes are well motivated - Could consider setting the interpolation method as an argument as well.

ml4gw/transforms/qtransform.py
```diff
@@ -130,15 +127,10 @@ def forward(self, fseries: torch.Tensor, norm: str = "median"):
         if norm:
             norm = norm.lower() if isinstance(norm, str) else norm
             if norm == "median":
-                # I swear there's a better way to do this
-                medians = torch.quantile(energy, q=0.5, dim=-1)
-                medians = medians.repeat(energy.shape[-1], 1, 1)
-                medians = medians.transpose(0, -1).transpose(0, 1)
+                medians = torch.quantile(energy, q=0.5, dim=-1, keepdim=True)
```
EthanMarx, line 130: Any reason not to use `torch.median`?

wbenoit26: Yeah, `torch.median` has different behavior than `numpy.median` for even-length arrays. It always takes the lower value, rather than the average of the two middle values.

wbenoit26:
The interpolation method was switched to by more in line with what `gwpy` does. You're right though, would be nice to have that be an argument. Let me try that out.

wbenoit26:
I've realized that this PR doesn't really work because torch's interpolation method is limited to 5 dimensions, and it expects the first two dimensions of the input to be the batch and channel dimension. Still, better to do the normalization step this way, I just need to revert some documentation.

wbenoit26:
@EthanMarx FYI, accidentally included a stray commit from a different change I was working on, which re-parameterizes the way the spectrogram shape is passed. Not a major update, though I see that I need to correct the type on the new variable.

EthanMarx:
All good. Maybe open up an issue if it's not gonna be resolved promptly

PR #119, opened 2024-05-08
`InMemoryDataset` improvements

### LINKED ISSUE #69: `InMemoryDataset` improvements
- Inherit from `torch.utils.data.IterableDataset` so that it can be picked up properly by torch's `DataLoader` class
- Implement iteration as a generator under `__iter__`
- Decide whether it makes sense to include a `y` timeseries separately and what assumptions we should make about it

EthanMarx:
Close #69 for use in `Amplfi`

### DIFF
--- a/ml4gw/dataloading/in_memory_dataset.py
+++ b/ml4gw/dataloading/in_memory_dataset.py
@@ -155,7 +155,7 @@ class InMemoryDataset(torch.utils.data.IterableDataset):
             # support it
             num_kernels = self.num_kernels ** len(self.X)
             return (num_kernels - 1) // self.batch_size + 1
-        
+
     def init_indices(self):
         """
         Initialize arrays of indices we'll use to slice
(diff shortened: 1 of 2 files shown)


ml4gw/dataloading/in_memory_dataset.py
```diff
@@ -204,36 +203,24 @@ def __iter__(self):
...
+            slc = slice(i * self.batch_size, (i + 1) * self.batch_size)
+            idx = indices[slc] * self.stride
+
+            # slice our timeseries
+            X = slice_kernels(self.X, idx, self.kernel_size)
+            if self.y is not None:
+                y = slice_kernels(self.y, idx, self.kernel_size)
+
+            if self.y is not None:
+                yield X, y
```
wbenoit26, line 225: Could move this statement into the previous conditional, but it doesn't really matter

wbenoit26 (changes requested):
Oh, and right, of course tests need to be changed to deal with the difference in return format

EthanMarx:
@wbenoit26 One more once over from you than should be good to go

wbenoit26 (approved):

PR #130, opened 2024-06-06
Added ringdown waveform function

tblodg23:


ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        times = torch.arange(num, dtype=torch.float64) / sample_rate
+        times -= duration / 2.0
+
+        self.register_buffer("times", times)
+
+    def __call__(
+        self,
+        frequency: ScalarTensor,
+        quality: ScalarTensor,
+        amplitude: ScalarTensor,
```
wbenoit26, line 29: We can calculate the amplitude from the other parameters based on equations 3.16, 3.6, and 3.7 from the thesis.

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        phase: ScalarTensor,
+        inclination: ScalarTensor,
+        #distance: ScalarTensor,
```
wbenoit26, line 32: Not sure why equations 7.1 and 7.2 don't have the distance in them, but they definitely should both be divided by `r`. Maybe that gets taken into account somewhere else in the original program, but we'll want to have it here.

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+            quality:
+                Quality factor of the ringdown waveform
+            amplitude:
+                Initial amplitude of the ringdown waveform
+            phase:
+                Initial phase of the ringdown waveform
+            inclination:
+                Inclination angle of the source.
+            distance:
+                Distance to the source.
```
wbenoit26, line 49: Should specify units. Likely, Mpc will be the most convenient

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        # ensure all inputs are on the same device
+        pi = torch.tensor([torch.pi], device=frequency.device)
+
+        # calculate cosines with inclination
+        cos_i = torch.cos(inclination)
+        cos_i2 = cos_i**2
+        sin_i = torch.sin(inclination)
+
+
+        # computing (A/r) times wave function h(t)
```
wbenoit26, line 71: This set of comments can be removed

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        # Original implementation based on equations 3.12 and 3.13
+        # h0 = (amplitude / distance) * torch.exp(-pi * frequency * self.times / quality) * torch.cos(2 * pi * frequency * self.times - phase)
+        # h_plus = (1 + cos_i2) * h0
+        # h_cross = 2 * cos_i * h0
+
+        # New implementation based on equations 7.5 and 7.6
+        a_plus = amplitude * (1 + cos_i2) * torch.exp(-pi * frequency * self.times / quality)
```
wbenoit26, line 78: For a little extra efficiency, you can calculate `torch.exp(-pi * frequency * self.times / quality)` just once and assign it to a variable to use in these lines

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        a_cross = amplitude * (2 * sin_i) * torch.exp(-pi * frequency * self.times / quality)
+
+        h_plus = a_plus * torch.cos(2 * pi * frequency * self.times + phase)
```
wbenoit26, line 81: Similarly, can define a variable for `2 * pi * frequency * self.times + phase`

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,90 @@
...
+        h_cross = a_cross * torch.sin(2 * pi * frequency * self.times + phase)
+
+        # ensure the dtype is double
+        h_plus = h_plus.double()
+        h_cross = h_cross.double()
+
+        return h_cross, h_plus
```
wbenoit26, line 88: We should add some tests (like [these](https://github.com/ML4GW/ml4gw/blob/main/tests/waveforms/test_sine_gaussian.py)) to make sure that we agree with established results. I need to do a little digging to figure out which function we should be comparing against, but you can try setting up the shell of the test script.

wbenoit26 (changes requested):

ml4gw/waveforms/ringdown.py
```diff
@@ -54,37 +58,43 @@ def __call__(
...
 
+        # Calculate spin 
+        spin = 1 - (2 / quality)**(20 / 9)
+        # Calculate mass 
+        mass = (1 / (2 * pi)) * (c**3 / (G * frequency)) * (1 - 0.63 * (2 / quality)**(2 / 3))
+
+        # Calculate amplitude 
+        F_Q = 1 + (1.34 / quality**2)
+        g_a = (1 + 0.1 * spin**2)
+        amplitude = np.sqrt(5 / 2) * epsilon * (6.67430e-11 * mass / (3e8**2)) * quality**(-0.5) * F_Q**(-0.5) * g_a**(-0.5)
```
wbenoit26, line 79: Need to switch to using `c` and `G` here

wbenoit26: Also, should have mentioned this earlier: we use a tool called `pre-commit` to automatically keep code tidy. In the top-level of the repo, can you run `poetry run pre-commit install`? Then, each time you make a `git commit`, it will check over your code, halt the commit if it finds any problems, and try to reformat it.

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,100 @@
...
+        pi = torch.tensor([torch.pi], device=frequency.device)
+
+        # Calculate spin 
+        spin = 1 - (2 / quality)**(20 / 9)
+        # Calculate mass 
+        mass = (1 / (2 * pi)) * (c**3 / (G * frequency)) * (1 - 0.63 * (2 / quality)**(2 / 3))
+
+        # Calculate amplitude 
+        F_Q = 1 + (1.34 / quality**2)
+        g_a = (1 + 0.1 * spin**2)
```
wbenoit26, line 78: Oh, also, these don't match equations after 3.16

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,112 @@
+import numpy as np
+import torch
+
+from ml4gw.types import ScalarTensor
+
+c = 299792458.0  # speed of ligt
+G = 6.67430e-11  # gravitational constant
```
wbenoit26, line 7: Ravi just recently added a [constants.py](https://github.com/ML4GW/ml4gw/blob/dev/ml4gw/constants.py) file that you can import these values from (e.g. `from ..constants import C, G`). For consistency, let's switch to using that file

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,112 @@
...
+        # add dimension for calculating waveforms in batch
+        frequency = frequency.view(-1, 1)
+        quality = quality.view(-1, 1)
+        epsilon = epsilon.view(-1, 1)
+        phase = phase.view(-1, 1)
+        inclination = inclination.view(-1, 1)
+        distance = distance.view(-1, 1)
+
+        # convert Mpc to m
+        m_per_Mpc = 3.0856775814913673e22
```
wbenoit26, line 67: There's also a value in `constants.py` for this that you can import

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,112 @@
...
+        Generate ringdown waveform based on the damped sinusoid equation.
+
+        Args:
+            frequency:
+                Central frequency of the ringdown waveform in Hz
+            quality:
+                Quality factor of the ringdown waveform
+            epsilon:
+                Fraction of black hole's mass radiated as gravitational waves
+            phase (rad):
```
wbenoit26, line 48: Should remove the " (rad)" from this line

ml4gw/waveforms/ringdown.py
```diff
@@ -0,0 +1,109 @@
+import constants
```
wbenoit26, line 1: This needs to be `..constants`

wbenoit26 (approved):

PR #131, opened 2024-06-10
fix prefactor_four_thirds in inspiral amplitude

ravioli1369:
The prefactor four_thirds variable has a `35371056 * eta2 * PI ** (4.0 / 3.0)` towards the bottom, whereas it should be `35371056 * eta2 )* PI ** (4.0 / 3.0)` according to LAL.

Please refer line 330 of https://lscsoft.docs.ligo.org/lalsuite/lalsimulation/_l_a_l_sim_i_m_r_phenom_d__internals_8c_source.html

This did not impact the waveform drastically, and I found all tests to be passing with roughly the same tolerance, so I've left the tests unchanged.

### DIFF
--- a/ml4gw/waveforms/phenom_d.py
+++ b/ml4gw/waveforms/phenom_d.py
@@ -955,9 +955,7 @@ def phenom_d_mrd_phase(Mf, eta, eta2, chi1, chi2, xi):
     mrd_phasing += (4.0 / 3.0) * (Mf.t() ** (3.0 / 4.0) * alpha3).t()
     mrd_phasing += (torch.atan(f_minus_alpha5_fRD.t() / fDM) * alpha4).t()
 
-    mrd_Dphasing = (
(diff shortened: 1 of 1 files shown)


wbenoit26:
Hi @ravioli1369, thanks for making this PR. Could you change it so that it's against the `dev` branch instead of the `main` branch? Other than that it seems good to go.

deepchatterjeeligo:
Thank you for catching this @ravioli1369. Could you make the merge request against `dev` branch locally? Otherwise this is good to go.

ravioli1369:
Looks like this is adding many extra commits, I'm not sure how to fix this, so should I just delete the branch, then branch it off of dev, make the change and open another PR?

wbenoit26:
@ravioli1369 Nice, thanks for fixing. One more minor request (which we should really have in the README). Can you run the following within your clone of the repo:
```
poetry run pre-commit install
poetry run pre-commit run --all
```

And then commit and push? The pre-commit code just keeps everything tidy, and once it's installed it will run automatically when you try to commit.

PR #136, opened 2024-06-24
Refactor waveforms + IMRPhenomPv2

ravioli1369:
This PR add the following: 
- Refactor the existing TaylorF2, PhenomD waveforms into `torch.nn.Module` classes, making use of `self.register_buffer` to avoid tensor initializations on the CPU.
- The PhenomD has been made into a subclass of TaylorF2 in order to use its functions without too many imports. 
- Addition of a constants.py file to keep everything in one place
- Fixing the TaylorF2 phase term used in PhenomD by subtracting off its 3PN term. Refer https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/lib/LALSimIMRPhenomD.c#L397-398. 
- Addition of the IMRPhenomPv2 waveform implementation in torch. This waveform is subclassed from PhenomD, again for easy use of PhenomD and TaylorF2 functions.

### DIFF
--- a/ml4gw/constants.py
+++ b/ml4gw/constants.py
@@ -26,6 +26,9 @@ gt = G * MSUN / (C**3.0)
 G MSUN / C^3 in seconds
 """
 
+MTSUN_SI = 4.925490947641266978197229498498379006e-6
+"""1 solar mass in seconds. Same value as lal.MTSUN_SI"""
+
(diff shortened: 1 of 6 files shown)


deepchatterjeeligo:
closing in favor of #137 and #138

PR #137, opened 2024-06-24
Refactor TaylorF2 and IMRPhenomD

ravioli1369:
This PR add the following: 
- Refactor the existing TaylorF2, PhenomD waveforms into `torch.nn.Module` classes, making use of `self.register_buffer` to avoid tensor initializations on the CPU.
- The PhenomD has been made into a subclass of TaylorF2 in order to use its functions without too many imports. 
- Addition of a constants.py file to keep everything in one place
- Fixing the TaylorF2 phase term used in PhenomD by subtracting off its 3PN term. Refer https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/lib/LALSimIMRPhenomD.c#L397-398.

### DIFF
--- a/ml4gw/waveforms/phenom_d.py
+++ b/ml4gw/waveforms/phenom_d.py
@@ -1,1303 +1,127 @@
 import torch
+from .taylorf2 import TaylorF2
 from torchtyping import TensorType
+from ..constants import MTSUN_SI, PI
+from .phenom_d_data import QNMData_a, QNMData_fdamp, QNMData_fring
 
-from . import phenom_d_data
-from .taylorf2 import MTSUN_SI, PI, taylorf2_amplitude, taylorf2_phase
 
+class IMRPhenomD(TaylorF2):
+    def __init__(self):
+        super().__init__()
+        self.register_buffer("QNMData_a", QNMData_a)
+        self.register_buffer("QNMData_fdamp", QNMData_fdamp)
+        self.register_buffer("QNMData_fring", QNMData_fring)
 
-# Utility functions taken from PhenomD utilities in lalsimulation
-# https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/lib/LALSimIMRPhenomD_internals.c
-def sigma1Fit(eta, eta2, xi):
-    return (
-        2096.551999295543
-        + 1463.7493168261553 * eta
-        + (
-            1312.5493286098522
-            + 18307.330017082117 * eta
-            - 43534.1440746107 * eta2
-            + (
-                -833.2889543511114
-                + 32047.31997183187 * eta
-                - 108609.45037520859 * eta2
-            )
-            * xi
-            + (
-                452.25136398112204
-                + 8353.439546391714 * eta
-                - 44531.3250037322 * eta2
-            )
-            * xi
-            * xi
+    def forward(
+        self,
+        f: TensorType,
+        chirp_mass: TensorType,
+        mass_ratio: TensorType,
+        chi1: TensorType,
+        chi2: TensorType,
+        distance: TensorType,
+        phic: TensorType,
+        inclination: TensorType,
+        f_ref: float,
+    ):
+        """
+        IMRPhenomD waveform
+
+        Returns:
+        --------
+        hp, hc
+        """
+        # shape assumed (n_batch, params)
+        if (
+            chirp_mass.shape[0] != mass_ratio.shape[0]
+            or mass_ratio.shape[0] != chi1.shape[0]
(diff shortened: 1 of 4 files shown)


ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+        ) * (
+            gamma2 <= 1
+        ).to(
+            torch.int
+        )
+        return res
+
+    def _linear_interp_finspin(self, finspin):
+        # chi is a batch of final spins i.e. torch.Size([n])
+        right_spin_idx = torch.bucketize(finspin, QNMData_a)
```
deepchatterjeeligo, line 584:
```suggestion
        right_spin_idx = torch.bucketize(finspin, self.QNMData_a)
```

ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+        right_spin_val = QNMData_a[right_spin_idx]
```
deepchatterjeeligo, line 585:
```suggestion
        right_spin_val = self.QNMData_a[right_spin_idx]
```

ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+        # QNMData_a is sorted, hence take the previous index
+        left_spin_idx = right_spin_idx - 1
+        left_spin_val = QNMData_a[left_spin_idx]
```
deepchatterjeeligo, line 588:
```suggestion
        left_spin_val = self.QNMData_a[left_spin_idx]
```

ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+
+        if not torch.all(left_spin_val < right_spin_val):
+            raise RuntimeError(
+                "Left value in grid should be greater than right. "
+                "Maybe be caused for extremal spin values."
             )
-            * xi
+        left_fring = QNMData_fring[left_spin_idx]
```
deepchatterjeeligo, line 595:
```suggestion
        left_fring = self.QNMData_fring[left_spin_idx]
```

ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+        right_fring = QNMData_fring[right_spin_idx]
```
deepchatterjeeligo, line 596:
```suggestion
        right_fring = self.QNMData_fring[right_spin_idx]
```

ml4gw/waveforms/phenom_d.py
```diff
@@ -1,1362 +1,1419 @@
...
+        slope_fring = right_fring - left_fring
+        slope_fring /= right_spin_val - left_spin_val
+
+        left_fdamp = QNMData_fdamp[left_spin_idx]
+        right_fdamp = QNMData_fdamp[right_spin_idx]
```
deepchatterjeeligo, line 601:
```suggestion
        left_fdamp = self.QNMData_fdamp[left_spin_idx]
        right_fdamp = self.QNMData_fdamp[right_spin_idx]
```

deepchatterjeeligo (changes requested):
Overall looks good @ravioli1369 There are a few instance where the phenom_d data is not referenced correctly. But that's an easy fix. Let's name the data attributes using lower case when registering buffers i.e. `qnm_data_a`, `qnm_data_fring`, `qnm_data_fring` etc.

Finally, could you also change the signature for the `TaylorF2` to take in `chirp_mass`, `mass_ratio` as the first two arguments, consistent with the IMR waveforms?

ravioli1369:
@deepchatterjeeligo so, apparently it's not as simple as doing a self.qnmdata to access the register buffers, there seems to be a `self.buffers` and `self.named_buffers` method to iterate over the buffers. 
@EthanMarx could you confirm if this is the right way of accessing them? When I had used self.qnmdata_, I got an error saying that parameter is not a part of my class.

EthanMarx:
@ravioli1369 When you register a buffer it should be accessible as a Class variable. Can you post the error

deepchatterjeeligo:
I agree with Ethan. E.g.
```
>>> class Net(torch.nn.Module):
...     def __init__(self):
...         super().__init__()
...         self.register_buffer('mean', torch.as_tensor(1.0))
...     def forward(self, x):
...         return x
... 
>>> net = Net()
>>> net.mean
tensor(1.)
```
So this should work ` self.register_buffer('qnm_data_a', QNMData_a)`?

deepchatterjeeligo (approved):
Looks good. Thanks @ravioli1369

PR #138, opened 2024-06-25
IMRPhenomP

ravioli1369:
This PR adds the IMRPhenom P implementation in torch.
This also fixes an error in Mass 1 and Mass 2 calculations for the TaylorF2 waveform

### DIFF
--- a/ml4gw/waveforms/phenom_p.py
+++ b/ml4gw/waveforms/phenom_p.py
@@ -32,8 +32,8 @@ class IMRPhenomPv2(IMRPhenomD):
         m1 must be larger than m2.
         """
 
-        m1 = chirp_mass * (1.0 + mass_ratio) ** 0.2 / mass_ratio**0.6
-        m2 = m1 * mass_ratio
+        m2 = chirp_mass * (1.0 + mass_ratio) ** 0.2 / mass_ratio**0.6
+        m1 = m2 * mass_ratio
 
         # # flip m1 m2. For some reason LAL uses this convention for PhenomPv2
         m1, m2 = m2, m1
--- a/ml4gw/waveforms/phenom_p.py
+++ b/ml4gw/waveforms/phenom_p.py
@@ -14,8 +14,8 @@ class IMRPhenomPv2(IMRPhenomD):
     def forward(
         self,
         fs: TensorType,
-        m1: TensorType,
-        m2: TensorType,
+        chirp_mass: TensorType,
+        mass_ratio: TensorType,
         s1x: TensorType,
         s1y: TensorType,
         s1z: TensorType,
@@ -31,6 +31,9 @@ class IMRPhenomPv2(IMRPhenomD):
         """
         m1 must be larger than m2.
         """
+
+        m1 = chirp_mass * (1.0 + mass_ratio) ** 0.2 / mass_ratio**0.6
+        m2 = m1 * mass_ratio
 
         # # flip m1 m2. For some reason LAL uses this convention for PhenomPv2
         m1, m2 = m2, m1
--- a/ml4gw/waveforms/phenom_p.py
+++ b/ml4gw/waveforms/phenom_p.py
@@ -28,7 +28,7 @@ class IMRPhenomPv2(IMRPhenomD):
         tc: TensorType,
         phiRef: TensorType,
         incl: TensorType,
-        f_ref: TensorType,
+        f_ref: float,
     ):
         """
         m1 must be larger than m2.
@@ -95,7 +95,7 @@ class IMRPhenomPv2(IMRPhenomD):
         Y20 = self.SpinWeightedY(thetaJN, 0, -2, 2, -0)
         Y21 = self.SpinWeightedY(thetaJN, 0, -2, 2, 1)
         Y22 = self.SpinWeightedY(thetaJN, 0, -2, 2, 2)
-        Y2 = torch.tensor([Y2m2, Y2m1, Y20, Y21, Y22])
+        Y2 = torch.stack((Y2m2, Y2m1, Y20, Y21, Y22))
 
         hPhenomDs, Dphase = self.PhenomPOneFrequency(
             fs,
@@ -133,12 +133,14 @@ class IMRPhenomPv2(IMRPhenomD):
             2 * PI * fs * (t0)
         )
         M_s = (m1 + m2) * MTSUN_SI
-        phase_corr_tc = torch.exp(-1j * fs * M_s * tc)
+        phase_corr_tc = torch.exp(
+            -1j * fs * M_s.unsqueeze(1) * tc.unsqueeze(1)
+        )
         hp *= phase_corr * phase_corr_tc
         hc *= phase_corr * phase_corr_tc
 
-        c2z = torch.cos(2 * zeta_polariz)
-        s2z = torch.sin(2 * zeta_polariz)
+        c2z = torch.cos(2 * zeta_polariz).unsqueeze(1)
+        s2z = torch.sin(2 * zeta_polariz).unsqueeze(1)
         hplus = c2z * hp + s2z * hc
         hcross = c2z * hc - s2z * hp
         return hplus, hcross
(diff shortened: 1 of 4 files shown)


ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        s2y: TensorType,
+        s2z: TensorType,
+        dist_mpc: TensorType,
+        tc: TensorType,
+        phiRef: TensorType,
+        incl: TensorType,
+        f_ref: float,
+    ):
+        """
+        m1 must be larger than m2.
```
deepchatterjeeligo, line 32: Let's add a good docstring here

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+
+class IMRPhenomPv2(IMRPhenomD):
+    def __init__(self):
+        super().__init__()
+
+    def forward(
+        self,
+        fs: TensorType,
+        m1: TensorType,
+        m2: TensorType,
```
deepchatterjeeligo, line 18: Like the rest of the waveforms, let's use chirp_mass, mass_ratio consistently.

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        """
+        original_shape = x.shape
+        x = x.flatten()
+        xp = xp.flatten()
+        fp = fp.flatten()
+
+        m = (fp[1:] - fp[:-1]) / (xp[1:] - xp[:-1])  # slope
+        b = fp[:-1] - (m * xp[:-1])
+
+        indicies = torch.searchsorted(xp, x, right=False) - 1
```
deepchatterjeeligo, line 332: typo

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+            - alphaoffset
+        ).mT
+
+        epsilon = (
+            (
+                angcoeffs["epsiloncoeff1"] / omega.T
+                + angcoeffs["epsiloncoeff2"] / omega_cbrt2.T
+                + angcoeffs["epsiloncoeff3"] / omega_cbrt.T
+                + angcoeffs["epsiloncoeff4"] * logomega.T
+                + angcoeffs["epsiloncoeff5"] * omega_cbrt.T
```
deepchatterjeeligo, line 195: Torch will [not support](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.T) the `.T` subsequently. Can you check if the `.mT`/`.t()` can be used as a replacement here and in other places?

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        return chi1_l, chi2_l, chip, thetaJN, alpha0, phi_aligned, zeta_polariz
+
+    # TODO: add input and output types
+    def SpinWeightedY(self, theta, phi, s, l, m):  # noqa: E741
+        "copied from SphericalHarmonics.c in LAL"
+        if s == -2:
+            if l == 2:  # noqa: E741
+                if m == -2:
+                    fac = (
+                        torch.sqrt(torch.tensor(5.0 / (64.0 * PI)))
```
deepchatterjeeligo, line 506: Have you checked if this works when the IMRPhenomPv2 instance is sent to the GPU? The `torch.tensor(5.0 / (64.0 * PI)` here and other similar instances will define the quantity on the CPU, hence asking.

deepchatterjeeligo: Also generally if we can avoid the `if...else` conditionals that would be even better. Can we register the Ylm factors as buffers and evaluate the Ylms as needed?

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        m1: TensorType,
+        m2: TensorType,
+        chi1_l: TensorType,
+        chi2_l: TensorType,
+        chip: TensorType,
+    ) -> TensorType:
+        M = m1 + m2
+        eta = m1 * m2 / (M * M)
+        eta2 = eta * eta
+        q_factor = torch.where(m1 >= m2, m1 / M, m2 / M)
```
deepchatterjeeligo, line 692: Can we avoid the `.where` if we switch to using chirp_mass and mass_ratio?

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        Amp0 = self.get_Amp0(Mf, eta)
+        dist_s = dist_mpc * MPC_SEC
+        Amp = ((Amp0 * Amp).T * (M_s**2.0) / dist_s).mT
+        # phase -= 2. * phic; # line 1316 ???
+        hPhenom = Amp * (torch.exp(-1j * phase))
+
+        linspace = torch.linspace(0.5, 1.5, 101)
+        fRDs = torch.outer(fRD, linspace)
+        delta_fRds = torch.median(torch.diff(fRDs, axis=1), axis=1)[0]
+        MfRDs = torch.zeros(*fRDs.shape)
```
deepchatterjeeligo, line 286: use `zeros_like` otherwise `MfRDs` is defined on the CPU

ml4gw/waveforms/phenom_p.py
```diff
@@ -0,0 +1,731 @@
...
+        Amp = self.phenom_d_amp(
+            Mf, m1, m2, eta, eta2, Seta, chi1, chi2, chi12, chi22, xi, dist_mpc
+        )[0]
+        Amp0 = self.get_Amp0(Mf, eta)
+        dist_s = dist_mpc * MPC_SEC
+        Amp = ((Amp0 * Amp).T * (M_s**2.0) / dist_s).mT
+        # phase -= 2. * phic; # line 1316 ???
+        hPhenom = Amp * (torch.exp(-1j * phase))
+
+        linspace = torch.linspace(0.5, 1.5, 101)
```
deepchatterjeeligo, line 283: Let's use a variable name that describes the quantity. And like the comment below, this defines the quantity on the CPU which might raise an error when `fRD` is a tensor on the GPU. Have you tested this?

deepchatterjeeligo (changes requested):
Hi @ravioli1369 we have the content, however, there are some places where I think certain quantities get defined on the CPU at runtime. When you create an instance of Pv2, send it to the GPU, and then pass the parameters and frequencies that live on the GPU, I suspect there may be some errors w.r.t. device compatibility. If you can test this case out. Besides this, there are some cosmetic changes w.r.t. docs, variable naming and such which should be easy changes.

ml4gw/waveforms/phenom_p.py
ravioli1369: this is my bad, the [ripple version](https://github.com/tedwards2412/ripple/blob/c451e87cd3130286fb2bd32c4148507ea05a3dff/src/ripplegw/waveforms/IMRPhenomPv2_utils.py#L354), which this code is based on assumes a convention of m1 being greater than m2. I copied this function (FinalSpin_inplane) from the LAL source code, which actually flips the variables if m1<m2. during my checks of the waveform, I found it didnt give correct results for m2>m1 due to other m1>m2 assumptions from the ripple code. I've made it such that the code doesn't decide values based on m1 and m2 now.

ml4gw/waveforms/phenom_p.py
ravioli1369: I dont think theres any way around it, since the factor `fac` depends on thetaJ, which depends on the input params.

ml4gw/waveforms/phenom_p.py
ravioli1369: I think it would be good to update the docstrings of all the other waveforms as well, by also mentioning the units of the parameters. I did face some issues when running phenomD, not realizing the distance is in Mpc. 

Maybe I could update them all at once in a new PR?

ml4gw/waveforms/phenom_p.py
ravioli1369: I used that to create a grid of frequencies around the ring down frequency for accurate differentiation. It's used immediately in the next line and nowhere else, so I guess it's simpler to just eliminate it and use `torch.linspace` directly.

ml4gw/waveforms/phenom_p.py
deepchatterjeeligo: Sure. Let's do it after.

deepchatterjeeligo (approved):
LGTM.

PR #144, opened 2024-07-02
Add functionalityfor CCSN simulations.

AndyC80297:
Add Computing second-derivative quadrupole moment to the two polarizations function.


EthanMarx:
@AndyC80297 It looks like there are extra commits from previous PRs.

I imagine you branched from `main`, and opened the PR against dev. Could you reopen this, branching from the `dev` branch?

PR #148, opened 2024-07-04
Docstring for phenom_d, phenom_p, tf2

ravioli1369:


EthanMarx:
@ravioli1369 Thanks for this. For the docstrings, we use `Args:` intead of `Parameters:` , and don't use the `-----` demarkation. Take a look at another function for an example.

ravioli1369:
@EthanMarx changed as requested

EthanMarx (approved):

PR #152, opened 2024-07-20
Add functionality for 3-D CCSN simulations.

AndyC80297:
Add functionality to transform second-derivative quadrupole moment to the two polarizations function for 3-D simulation waveform.

### DIFF
--- a/ml4gw/gw.py
+++ b/ml4gw/gw.py
@@ -257,6 +257,55 @@ def compute_observed_strain(
         waveforms, theta, phi, detector_vertices, sample_rate
     )
 
+def pol_from_quad(
+    quads,
+    theta,
+    phi
+):
+
+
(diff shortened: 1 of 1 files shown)


wbenoit26:
This looks good. I've got a couple of changes to suggest:

1. Move this function into its own file in the [waveforms](https://github.com/ML4GW/ml4gw/tree/main/ml4gw/waveforms) directory and have it follow the same structure as the other files there. So, make a class that inherits from `torch.nn.Module` and have the `forward` method of this class be the function that creates the polarizations. That will make this PR more consistent with the existing structure.

2. If you haven't already, run `poetry install` in the root of this repo to install the environment, and then run `poetry run pre-commit install` to install the pre-commit hooks, which will structure and organize your code for you whenever you commit.

PR #153, opened 2024-07-31
Jaxtyping

### LINKED ISSUE #83: Move to `jaxtyping`
`torchtyping` now [strongly recommends](https://github.com/patrick-kidger/torchtyping#type-annotations-for-a-tensors-shape-dtype-names-) moving to a newer, more general library [`jaxtyping`](https://github.com/google/jaxtyping), which looks like it has Google's support. Despite the name, it has no required jax dependency and only appears to depend on numpy, typeguard, and typing_extensions. Seems like it implements some nice features as well, e.g. [modifiers](https://docs.kidger.site/jaxtyping/api/array/) to add more shape flexibility (e.g. when a batch axis is allowed but not required).

Note that in `torchtyping`'s comment they indicate that `jaxtyping` is compatible with static-type checkers, but the [FAQs](https://docs.kidger.site/jaxtyping/faq/) indicate that this support is limited at best, which is too bad because that would be pretty slick. But still worth transitioning to what seems like it will be the longer-term supported library.

wbenoit26:
Switching over from `torchtyping` to `jaxtyping`. Closes #83.

- [x] Replace all current instances of `torchtyping` to `jaxtyping`
- [x] Do better type-hinting throughout

### DIFF
--- a/ml4gw/gw.py
+++ b/ml4gw/gw.py
@@ -13,7 +13,8 @@ https://github.com/lscsoft/bilby/blob/master/bilby/gw/detector/interferometer.py
 from typing import List, Tuple, Union
 
 import torch
-from jaxtyping import Array, Float
+from jaxtyping import Float
+from torch import Tensor
 
 from ml4gw.types import (
     NetworkDetectorTensors,
@@ -67,7 +68,7 @@ def compute_antenna_responses(
     phi: ScalarTensor,
     detector_tensors: NetworkDetectorTensors,
     modes: List[str],
-) -> Float[Array, "batch polarizations num_ifos"]:
+) -> Float[Tensor, "batch polarizations num_ifos"]:
     """
     Compute the antenna pattern factors of a batch of
     waveforms as a function of the sky parameters of
@@ -197,7 +198,7 @@ def compute_observed_strain(
     detector_tensors: NetworkDetectorTensors,
     detector_vertices: NetworkVertices,
     sample_rate: float,
-    **polarizations: Float[Array, "batch time"],
+    **polarizations: Float[Tensor, "batch time"],
 ) -> WaveformTensor:
     """
     Compute the strain timeseries $h(t)$ observed by a network
@@ -289,8 +290,8 @@ def compute_ifo_snr(
     responses: WaveformTensor,
     psd: PSDTensor,
     sample_rate: float,
-    highpass: Union[float, Float[Array, " frequency"], None] = None,
-) -> Float[Array, "batch num_ifos"]:
+    highpass: Union[float, Float[Tensor, " frequency"], None] = None,
+) -> Float[Tensor, "batch num_ifos"]:
     r"""Compute the SNRs of a batch of interferometer responses
 
     Compute the signal to noise ratio (SNR) of individual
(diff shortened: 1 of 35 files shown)


wbenoit26:
@EthanMarx I think this is good for a look-over. `jaxtyping` isn't really the magic bullet we were hoping for. I think a lot of what we wanted to do with it, like conditional typing, is beyond what it's capable of. To be honest, I'm not sure that really adds much clarity.

Still, it's good to be using a library that will continue to have support, and I think we can iteratively improve things as we go.

ml4gw/gw.py
```diff
@@ -32,7 +33,7 @@
 # define some tensor shapes we'll reuse a bit
 # up front. Need to assign these variables so
 # that static linters don't give us name errors
-batch = num_ifos = polarizations = time = frequency = space = None  # noqa
+# batch = num_ifos = polarizations = time = frequency = space = None  # noqa
```
EthanMarx, line 36: do we need this comment anymore?

ml4gw/types.py
```diff
@@ -1,10 +1,25 @@
...
+NetworkVertices = Float[Tensor, "num_ifos 3"]
+NetworkDetectorTensors = Float[Tensor, "num_ifos 3 3"]
+
+
+TimeSeries1d = Float[Tensor, "time"]
+TimeSeries2d = Float[TimeSeries1d, "channel"]
+TimeSeries3d = Float[TimeSeries2d, "batch"]
+TimeSeries1to3d = Union[TimeSeries1d, TimeSeries2d, TimeSeries3d]
+
+FrequencySeries1d = Float[Tensor, "frequency"]
```
EthanMarx, line 20: Could all of these be succinctly put as 
`FrequencySeries = Float[Tensor, "*batch *channel frequency"]`
Where the `*` means it can refer to 0 or more dimensions

EthanMarx: https://docs.kidger.site/jaxtyping/api/array/

ml4gw/waveforms/phenom_p.py
```diff
@@ -412,25 +418,25 @@ def L2PNR(self, v: TensorType, eta: TensorType) -> TensorType:
...
         f_ref: float,
-        phiRef: TensorType,
-        incl: TensorType,
-        s1x: TensorType,
-        s1y: TensorType,
-        s1z: TensorType,
-        s2x: TensorType,
-        s2y: TensorType,
-        s2z: TensorType,
+        phiRef: Float[Tensor, ""],
```
EthanMarx, line 424: Would this be the same as the `ScalarTensor` in `ml4gw.types` ?

ml4gw/types.py
wbenoit26: The multiple axes symbols can only be used once per array

wbenoit26: One of the reasons I ended up being disappointed with `jaxtyping` - having that functionality would be pretty useful, but I found an old issue about it and it sounds like they deliberately don't want that to be possible

ml4gw/gw.py
wbenoit26: Nope, I'll remove it

ml4gw/waveforms/phenom_p.py
wbenoit26: Yeah, good catch. Also, some of the types in the waveforms files are definitely wrong - I must've done an overzealous find and replace. Let me get that fixed up.

ml4gw/types.py
wbenoit26: Oh, and doing `Float[Tensor, "*batch frequency"]` would effectively mean "at least 1d", which would be a problem for some functions that cap at 3d.

EthanMarx (approved):
I think this looks good!

PR #155, opened 2024-08-11
Add functionality for 3-D CCSN simulations.

AndyC80297:


wbenoit26 (approved):
This looks good to me. Is there an existing library that we could write tests to compare this against?

deepchatterjeeligo:
This seems like a pretty modular addition? Is this still under consideration?

wbenoit26:
For sure. I'd like at least some basic tests to go along with it, but this would be great to have in here.

PR #156, opened 2024-09-26
Allow python 3.12

wbenoit26:
Expands support to python 3.12 ~~, and restricts `rpds-py` to `<0.20` because poetry is having trouble for reasons I can't figure out.~~


wbenoit26:
@EthanMarx This should be good to go as well. Would be great if we could make a 0.5.1 release with this update now that we don't have `ml4gw` as a submodule in `aframe`. After that, I'll work on getting `aframe` up to date.

wbenoit26:
@EthanMarx Can you take a look at this and the `hermes` PR?

EthanMarx (approved):
Glad we only needed to bump the versions - will make a point release

PR #157, opened 2024-09-28
Add spline interpolation

wbenoit26:
Adding the spline interpolation into `ml4gw` and integrating it into the Q-transform. Still a lot to do, but I wanted to get something pushed so people could take a look at it. The main changes compared to the [original](https://github.com/dottormale/Qtransform_torch/blob/main/torch_spline_interpolation.py) are:

- Unification of 1D and 2D splines into a single class. I noticed that the 1D interpolation was for some reason a lot slower than the 2D, so I tried to generalize the 2D to handle both cases. There are some percent-level discrepancies between the two that I still need to track down though.
- Pre-calculating the basis matrices. Based on the parameters of the Q-transform, we know the input and output shapes of the interpolation, so we can save a little time by calculating these up front. I've included the option to specify different input and output coordinates after object initialization if desired, but for our use case, I think we generally won't need to do that.
- Allowing a channel dimension in addition to a batch dimension so that we can interpolate data from multiple detectors at the same time.
- Some formatting changes to bring things more in line with how things are set up in `ml4gw`

To-do:
- [x] Establish consistency with original code. For 2D, the results agree exactly, but I want to track down the difference for the 1D case.
- The difference between the 1D and 2D comes down to numerical precision. Even after casting everything to `double`, minor deviations accumulate from various factors: different data input shape, `einsum` vs. `matmul`, order of multiplication, etc. These deviations get exacerbated by the final `torch.linalg.solve` in calculating the spline coefficients. I don't think there's anything I can do about this, but the approach is valid and the typical error between the two methods is small, so I'm going to mark this as done.
- [x] Make a nicer/more intuitive API for `SplineInterpolate`. The current setup makes it a little awkward to do 1D interpolation~~,and I want to put back the functionality from the original code that allowed specifying just the input/output size~~.
- [x] Add data validation checks to `SplineInterpolate`
- [ ] Set up rigorous testing against `scipy`. I know that this interpolation has edge effects that aren't present in `scipy`, but it would be good to ensure that everything mostly agrees.
- [x] Improve integration with `SingleQTransform`. It would be nice to be able to specify whether the full spline interpolation is desired, or whether some preset `torch` option is sufficient. This is particularly important because the matrix inversion that occurs during the spline interpolation takes up a lot of memory for large matrices, and so may not be desirable  for some use cases.
- [x] Documentation
- [x] Add Q-transform tests with spline interpolation
- [x] There's possibly some more optimization to be done by using the fact that many of the Q-tiles have the same shape, so they can be stacked and interpolated as a group.
- [ ] Figure out what the memory constraints are

### DIFF
--- a/ml4gw/transforms/__init__.py
+++ b/ml4gw/transforms/__init__.py
@@ -4,6 +4,6 @@ from .scaler import ChannelWiseScaler
 from .snr_rescaler import SnrRescaler
 from .spectral import SpectralDensity
 from .spectrogram import MultiResolutionSpectrogram
-from .spline_interpolation import SplineInterpolation
+from .spline_interpolation import SplineInterpolate
 from .waveforms import WaveformProjector, WaveformSampler
 from .whitening import FixedWhiten, Whiten
--- a/ml4gw/transforms/__init__.py
+++ b/ml4gw/transforms/__init__.py
@@ -4,5 +4,6 @@ from .scaler import ChannelWiseScaler
 from .snr_rescaler import SnrRescaler
 from .spectral import SpectralDensity
 from .spectrogram import MultiResolutionSpectrogram
+from .spline_interpolation import SplineInterpolation
 from .waveforms import WaveformProjector, WaveformSampler
 from .whitening import FixedWhiten, Whiten
--- a/ml4gw/transforms/qtransform.py
+++ b/ml4gw/transforms/qtransform.py
@@ -6,6 +6,7 @@ import torch.nn.functional as F
 from jaxtyping import Float, Int
 from torch import Tensor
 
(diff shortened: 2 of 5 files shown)


wbenoit26:
@EthanMarx This is at the point that it would be good to have you take a look at it. There remain a couple of issues with matching exactly with SciPy's implementation, but I think the code is usable with the caveats that I describe in the documentation.

EthanMarx:
So the spline interpolation is a general interpolator that could be (potentially) used for different use cases, and then you are just implementing it for the Qtransform?

wbenoit26:
Exactly, yeah

EthanMarx (approved):

PR #162, opened 2024-11-05
draft: linear -> spline interp for waveform data

deepchatterjeeligo:
lalsimulation uses spline interpolation for the phenom data. We have been using linear interpolation due to lacking implementation which we have now.

### DIFF
--- a/tests/waveforms/test_cbc_waveforms.py
+++ b/tests/waveforms/test_cbc_waveforms.py
@@ -225,16 +225,16 @@ def test_phenom_d(
     hc_torch = hc_torch[torch_mask]
 
     assert np.allclose(
-        1e21 * hp_lal_data.real, 1e21 * hp_torch.real.numpy(), atol=2e-4
+        1e21 * hp_lal_data.real, 1e21 * hp_torch.real.numpy(), atol=3e-4
     )
     assert np.allclose(
-        1e21 * hp_lal_data.imag, 1e21 * hp_torch.imag.numpy(), atol=2e-4
+        1e21 * hp_lal_data.imag, 1e21 * hp_torch.imag.numpy(), atol=3e-4
     )
     assert np.allclose(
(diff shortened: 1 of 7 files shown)


EthanMarx:
@deepchatterjeeligo can you comment on the tolerance increases for the tetss? If we're supposedly matching lalinference better shouldn't those go down?

tests/waveforms/test_cbc_waveforms.py
```diff
@@ -225,16 +225,16 @@ def test_phenom_d(
     hc_torch = hc_torch[torch_mask]
 
     assert np.allclose(
-        1e21 * hp_lal_data.real, 1e21 * hp_torch.real.numpy(), atol=2e-4
+        1e21 * hp_lal_data.real, 1e21 * hp_torch.real.numpy(), atol=3e-4
```
deepchatterjeeligo, line 228: I could not make the tests pass without bumping the threshold a bit, which is strange given that spline interpolation should work better than linear. Any thought whether tweaking the kx, sx values can help? CC @wbenoit26

wbenoit26: Honestly, I might steer clear of using the spline interpolation for this application for now. There are significant deviations from scipy's interpolation if you interpolate too close to the edges of your input range. This is workable for something like the Q-transform, where we can crop out those effects, but that's not something that you can do here. I'd be curious to see if the match is better for values well within the range, though.

deepchatterjeeligo: OK. I'll leave this marked as draft until the sanity is met.

wbenoit26: Yeah, I think this would be good to revisit. There should be much better agreement now, though it would be good also to see how much this slows down the calculation.

wbenoit26:
@deepchatterjeeligo It looks like the current interpolation method is a barrier to being able to compile PhenomPv2. I'll be curious whether the SplineInterpolation works here, because I think that should be able to be compiled.

PR #163, opened 2024-11-14
Add multi-rate resampling

wbenoit26:
Adds a transform that takes a timeseries at a given sample rate and returns a timeseries with different segments at different sample rates.

Rather than splitting the timeseries into component segments and resampling each individually, this resamples the entire timeseries multiple times, splitting and concatenating to get the final result. This is less computationally efficient, but avoids having multiple resampled edges in the final product. Some kind of in-between solution is probably better, splitting the original timeseries into overlapping segments that get resampled and cropped.

Need to add testing as well.

### DIFF
--- a/ml4gw/transforms/__init__.py
+++ b/ml4gw/transforms/__init__.py
@@ -4,6 +4,6 @@ from .scaler import ChannelWiseScaler
 from .snr_rescaler import SnrRescaler
 from .spectral import SpectralDensity
 from .spectrogram import MultiResolutionSpectrogram
-from .spline_interpolation import SplineInterpolation
+from .spline_interpolation import SplineInterpolate
 from .waveforms import WaveformProjector, WaveformSampler
 from .whitening import FixedWhiten, Whiten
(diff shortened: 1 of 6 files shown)


EthanMarx:
This is great. Would be curious to 
1. See some visualizations of this with raw waveforms / get idea of how much data we can save
3. implement a prototype of this in aframe (even if its buggy) just to see if it's a bottleneck computationally

wbenoit26:
Definitely. We can play around a bit with what frequencies we go to/where the transition points are. Just messing around a bit, if we have a 5-5 system from 32 Hz, it takes about 5.5 seconds to merge, and has a final frequency of 438 Hz (at least, to whatever order `lalapps_chirplen` goes to). So, if we have a 6 second window with the coalescence time in the final quarter second, we can have breakpoints at `[4, 5, 5.5, 5.75]` with sample rates of `[128, 256, 512, 1024, 2048]` and keep the signal frequency below the Nyquist frequency of our sample rate the whole time.

This results in a window of size 2304, which is smaller than our current window. I did a little benchmarking and found that the resampler could process a tensor of shape `(2048, 2, 6 * 2048)` in 7 ms, which should be fast enough. I'll make some plots when I get chance later today.

PR #165, opened 2024-11-26
Specify output size for irfft

wbenoit26:
When the last dimension of `X` was odd, computing the rfft and irfft would result in losing a sample, which lead to unexpected behaviors downstream. Explicitly setting the output shape of the irfft fixes this.


EthanMarx (approved):
This should only affect odd inputs correct?

wbenoit26:
Yeah, that's correct

PR #167, opened 2024-12-10
Updates to `IMRPhenomP` api

EthanMarx:
- Make  `IMRPhenomP` api consistent with `IMRPhenomD`
- `tc` defaults to batch of zeros if not specified


deepchatterjeeligo (approved):
looks good

PR #168, opened 2024-12-12
add precessing spin conversion

EthanMarx:
Converts from bilby to lalsim parameters.


EthanMarx:
@deepchatterjeeligo I need to clean up the documentation, but wanted to get your feedback on this. 

Specifically, I don't think this should this live in the top level `ml4gw`. Im thinking of restructuring the `waveforms` module a bit. What do you think?

Also, as far as parameter naming conventions, I was thinking of using bilby's as opposed to `lalsims`. For example, bilby calls the spin tilt angles `tilt_{1,2}` where as lalsim calls them `theta_{1,2}`.  I think being consistent with bilby is more important. 

@wbenoit26 I think this will allow us to remove dependency on `lalsim` and that for loop in the `ledger` library.

wbenoit26:
Nice. We don't have an explicit dependency on `lalsim` in `ledger`, it just comes in through `pycbc`. But it'll be great to have this in `ml4gw`. Looks like there's a file you haven't pushed though.

deepchatterjeeligo (approved):
This is convincing test. I was not aware of the lalsimulation API, but I took a look and it looks like the right thing to benchmark against.

EthanMarx:
Ah yeah I forgot the actual file lol

ml4gw/conversion.py
```diff
@@ -0,0 +1,177 @@
...
+        theta1:
+        theta2:
+        phi12:
+        chi1:
+        chi2:
+        mass_1:
+    """
+
+    # check if f_ref is valid
+    if f_ref == 0.0:
```
wbenoit26, line 59: Might as well check for `f_ref <= 0.0` here

wbenoit26 (commented):
I left a few small comments, and you mentioned that you were going to complete the documentation. I think this looks good though - seems like a fairly one-to-one translation from lalsim? It'll be nice to be able to replace the loop in Aframe with this

EthanMarx:
Addressed your comments @wbenoit26. 

I took the liberty of organizing the waveforms module a bit. Just added `cbc` and `adhoc` submodules.

EthanMarx:
@deepchatterjeeligo Can I get one more lookover?

deepchatterjeeligo (approved):
Looks good at first glance. Good that we have benchmarks with lalsim.

deepchatterjeeligo:
Maybe rename adhoc with unmodeled?

PR #172, opened 2024-12-16
Fix bug in `IMRPhenomD` fmax calc

EthanMarx:
- Fixes bug in `IMRPhenomD` fmax calc that produced nans
- Makes waveform tests more robust by using wider distributions and more samples


deepchatterjeeligo (approved):
This looks good.

PR #176, opened 2024-12-29
IMRPhenomPv2 fix

ravioli1369:


ml4gw/waveforms/cbc/phenom_d.py
```diff
@@ -384,17 +407,20 @@ def phenom_d_inspiral_amp(
 
         return amp, Damp
 
-    def phenom_d_phase(self, Mf, mass_1, mass_2, eta, eta2, chi1, chi2, xi):
+    def phenom_d_phase(
+        self, Mf, mass_1, mass_2, eta, eta2, chi1, chi2, xi, fRD=None, fDM=None
```
deepchatterjeeligo, line 411: Looks good. Just leave a comment here that the `None` values are for phenomd and for phenomp the value need to be supplied.

ml4gw/waveforms/cbc/phenom_d.py
```diff
@@ -454,15 +480,18 @@ def phenom_d_phase(self, Mf, mass_1, mass_2, eta, eta2, chi1, chi2, xi):
...
+    ):
         alpha1 = self.alpha1Fit(eta, eta2, xi)
         alpha2 = self.alpha2Fit(eta, eta2, xi)
         alpha3 = self.alpha3Fit(eta, eta2, xi)
         alpha4 = self.alpha4Fit(eta, eta2, xi)
         alpha5 = self.alpha5Fit(eta, eta2, xi)
 
         # merger ringdown
-        fRD, fDM = self.fring_fdamp(eta, eta2, chi1, chi2)
+        if (fRD is None) or (fDM is None):
```
deepchatterjeeligo, line 493: For completeness, throw a RuntimeError if only one of fRD or fDM is not none.

ravioli1369: would be better to make the `or` into an `and`?

deepchatterjeeligo: Sure, but check the behavior. It's better to be explicit and to raise the error for unexpected combinations.

ravioli1369: yea I did that in the latest commit, could you take a look?

deepchatterjeeligo (approved):

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -409,14 +434,14 @@ def test_phenom_p():
         hc_lal_data = hc_lal.data.data[lal_mask]
 
         assert np.allclose(
-            1e21 * hp_lal_data.real, 1e21 * hp_ml4gw.real.numpy(), atol=2e-3
+            1e21 * hp_lal_data.real, 1e21 * hp_ml4gw.real.numpy(), atol=1e-2
```
wbenoit26, line 437: Why was it necessary to increase the tolerance? Do we agree less closely with lalsim after these changes? Or are we testing a wider range of parameter space?

ravioli1369: The previous version of the waveform didn't match at all with lalsim when the tests were expanded to a larger parameter space. Now we match well in the full space, but certain parameters cause differences larger than ~6e-3, so I made it 1e-2.

ml4gw/waveforms/cbc/phenom_p.py
```diff
@@ -320,42 +328,61 @@ def PhenomPOneFrequency(
...
         )
-        diffRDphase = -diffRDphase[:, 50]
-        # MfRD = torch.outer(M_s, fRD)
-        # Dphase = torch.diag(
-        #     -self.phenom_d_phase(
-        #         MfRD, m1, m2, eta, eta2, chi1, chi2, xi)[1] * M_s
-        # ).view(-1, 1)
+        # interpolate at x = 1, as thats the same as f = fRD
+        diffRDphase = -self.interpolate(
+            torch.tensor([1]), x[1:-1], diffRDphase
```
AndyC80297, line 384: The x[1:-1] and diffRDphase will have different shapes when the batch size is not one. 
Probably need to let x[1:-1] follow the shape of diffRDphase so the function can correctly interpolate.

EthanMarx: @ravioli1369 can you look into this?

EthanMarx: We should also add to the tests a test that uses batched inputs

deepchatterjeeligo: Ok this is a good catch. They TaylorF2 and PhenomD tests had the same parameters repeated along the batch dimension. I think that is the easiest extension to the tests to ensure the plumbing is working as expected.

deepchatterjeeligo: Sorry for the oversight during the review.

PR #180, opened 2025-01-06
Add demo notebook

wbenoit26:
This adds a demo notebook with instructions/examples for using some of the tools in `ml4gw`


EthanMarx:
This is great. What are your thoughts on using the tutorials as a starting point for building out the docs? ( we should also migrate to using `readthedocs` like our other projects)

wbenoit26:
I see the tutorials as complementary to documentation more than as a place to build from. I think this notebook is a useful place for people to get started on understanding how to use various modules and what the advantages are, but it doesn't really give much detail about the specifics, and I wouldn't want it to. What I think would be good is to create a `readthedocs` page, and then link to the relevant pages throughout the notebook.

EthanMarx:
All for keeping the notebook around in notebook form, especially for the end to end training example. This might just be a personal thing, but I'm much more likely to poke around a documentation page with code examples then to download a notebook and run it.

wbenoit26:
Oh I see, I hadn't thought about that option. Yeah, totally, when we build the documentation we should take some snippets from here (and if we put code in the documentation, we should consider adding it into the notebook as well, though I think that matters less).

wbenoit26:
@EthanMarx Can we merge this in?

EthanMarx (approved):

PR #182, opened 2025-01-07
Restrict scipy in env and for tox

wbenoit26:
Looks like `btdtri` was removed from SciPy's special functions in 1.15.0. `tox` needed the version restriction to be explicit.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,13 +13,13 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.5.0"
+version = "4.5.2"
 description = "High level compatibility layer for multiple asynchronous event loop implementations"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


deepchatterjeeligo:
Got it. They don't have a good migration guide but the functions seems to have been renamed (was looking through the scipy issues and changelog). So bilby needs to rename those with [scipy.special.betainc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.betainc.html#scipy.special.betainc) and [scipy.special.betaincinv](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.betainc.html#scipy.special.betaincinv).

deepchatterjeeligo:
But bilby seems to have that constrain [see here](https://git.ligo.org/lscsoft/bilby/-/blob/master/requirements.txt?ref_type=heads) no?

wbenoit26:
Bilby's restriction should work, but this has something to do with how dependency resolution occurs with tox. I probably didn't need the restriction in the dev dependencies section (the `poetry.lock` file was fine), but whatever happens in the installation setup before the tests ends up overwriting the desired scipy version.

PR #186, opened 2025-01-18
Update readme and add license and citation

wbenoit26:


EthanMarx (approved):
Looks awesome thanks for doing this

PR #191, opened 2025-01-30
Add `fnames_per_batch` argument to `HDF5Dataset`

EthanMarx:
Adds argument for specifying number of files to sample per batch. If left as `None` will default to all available fnames

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,7 @@ repos:
   - id: poetry-check
   - id: poetry-lock
     args: [--check, --no-update]
+- repo: https://github.com/dhruvmanila/remove-print-statements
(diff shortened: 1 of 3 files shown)


wbenoit26:
Would be good to have a check that `fnames_per_batch` isn't greater than `len(fnames)`

deepchatterjeeligo (approved):
Looks good. One (nitpick) comment: rather that `fnames_per_batch` use `n_files_per_batch` or `num_files_per_batch`. However, up to you.

EthanMarx:
Both comments addressed

wbenoit26:
You've got a few `print` statements in there - we should add that print removal pre-commit hook to this repo

EthanMarx:
Done - added that hook as well

PR #192, opened 2025-01-30
Increased waveform tolerance

wbenoit26:
Did some testing because our tests here were failing too often. I was seeing a failure about 1 / 2,500 waveform samples, and we do 2,400 waveforms in our CI, so it's not too surprising that we have so many failures. Doubling the tolerance reduces the rate to about 1 / 10,000, which is more reasonable.

I'd increase more, but some of the differences were greater than 1e-2, which I think would be too large of a tolerance. Here's a histogram of the maximum differences of 25,000 samples.

diff_hist


EthanMarx:
Thanks for this. Are these maximum differences on the `hp * 1e21`?

EthanMarx (approved):

wbenoit26:
Yes, that's correct. The largest I saw of the 50k was 0.097, which is very outside tolerance but seemingly pretty rare

PR #195, opened 2025-02-05
additional ad-hoc waveforms (gaussian, wnb, strings) 

eric-moreno:


deepchatterjeeligo:
@eric-moreno is this still under consideration. It looks like these waveforms exist in lalsimulation, could you add unittests against those?

deepchatterjeeligo (changes requested):
@eric-moreno please add unittest for the new waveforms.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,156 @@
...
+
+        dt = 1.0 / sample_rate
+        f_low = 1.0  # Hard-coded in the LAL code
+
+        # We want length = 2 * floor( (9.0/f_low)/dt/2 ) + 1,
+        # doing the arithmetic as GPU-friendly if desired.
+        # But to create an actual Python shape, we do .cpu().item() at the end.
+        temp = (9.0 / f_low) / dt / 2.0  # float in Python
+        # If you want GPU math, push it to a GPU tensor first:
+        temp_t = torch.tensor(temp, device=device, dtype=torch.float32)
```
deepchatterjeeligo: this seems to be a constant. Why not make it a `register_buffer`

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,156 @@
...
+        half_len = torch.floor(temp_t)  # GPU-based floor
+        half_len_int = int(half_len.cpu().item())  # bring back to Python int
```
deepchatterjeeligo: We should not have `.cpu()` in the waveform generator

ml4gw/waveforms/adhoc/cosmic_string.py
AndyC80297: Since we’re only calculating the required time-window size, using a torch.Tensor operation is overkill. I will change it to math.floor.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
+import math
+
+import torch
+from scipy.signal.windows import tukey
+from torch import Tensor
+
+from ml4gw.types import BatchTensor
+
+
+class GenerateString(torch.nn.Module):
```
wbenoit26, line 10: Let's call this `CosmicString` to be more in line with the other naming conventions.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+      - f_low = 1 Hz
+      - length = floor(9.0 / f_low / dt / 2) * 2 + 1  => ~9 seconds total
+      - The waveforms are built in frequency domain and iFFT to time domain.
+    """
+
+    def __init__(self, sample_rate: float, duration: float, device="cpu"):
+        """
+        Args:
+            sample_rate: sampling rate (Hz).
+            device: which device ("cpu" or "cuda") to store buffers on.
```
wbenoit26, line 24: Missing `duration` from the docstring

wbenoit26: Also, the `__init__` args should go in the class-level docstring.

ml4gw/waveforms/adhoc/white_noise_burst.py
```diff
@@ -0,0 +1,217 @@
+import math
+
+import torch
+from scipy.signal.windows import tukey
+from torch import nn
+
+from ml4gw.types import BatchTensor  # typically an alias for torch.Tensor
+
+torch.manual_seed(42)
```
wbenoit26, line 9: We shouldn't be setting a seed here, it will overwrite any seed that a user sets.

ml4gw/waveforms/adhoc/white_noise_burst.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def semi_major_minor_from_e(eccentricity):
```
wbenoit26, line 12: This looks identical to the function in `sine_gaussian.py`. I suggest we pull it out into a `adhoc/utils.py` file that both import from.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+            f_high: (batch,) freq above which we apply exponential taper.
+        Returns:
+            (h_cross, h_plus): shape (batch, self.length).
+            The cross polarization is zero (as in LAL).
+        """
+        # Reshape for batch dimension
+        amplitude = amplitude.view(-1, 1)
+
+        device = self.freq.device
+        # We'll assume amplitude is already on the same device
```
wbenoit26, line 98: I think this comment can be removed.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        super().__init__()
+        self.sample_rate = sample_rate
+
+        dt = 1.0 / sample_rate
+        f_low = 1.0  # Hard-coded in the LAL code
+
+        # We want length = 2 * floor( (9.0/f_low)/dt/2 ) + 1,
+        # doing the arithmetic as GPU-friendly if desired.
+        # But to create an actual Python shape,
```
wbenoit26, line 34: This comment seems out of date

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+        self.win_end = half_len_int + half_win_len
+
+        # Register these as buffers so they move with model.to(device)
+        self.register_buffer("freq", freq)  # shape (freq_bins,)
+        self.register_buffer(
+            "phase_factor", phase_factor
+        )  # shape (freq_bins,)
+
+        # Build final Tukey(0.5) window for time domain
+        tukey_window = tukey(M=length, alpha=0.5)
```
wbenoit26, line 72: For the given parameters, this Tukey window isn't doing anything. With `sim_duration=9`, `length=18433`. If `duration=4`, then the slice takes `[5120:13312]`, where the window is equal to one. So no smoothing happens.

wbenoit26: It looks like this is true in the other modules as well.

ml4gw/waveforms/adhoc/gaussian.py
```diff
@@ -0,0 +1,65 @@
...
+        self,
+        hrss: BatchTensor,
+        gaussian_width: BatchTensor,
+    ):
+        # The gaussian_width should be smaller than self.duration
+        hrss = hrss.view(-1, 1)
+        gaussian_width = gaussian_width.view(-1, 1)
+
+        # correct LAL normalisation
+        h0 = hrss / torch.sqrt(
```
wbenoit26, line 45: The polarizations are sliced after this normalization, so you don't get back the requested hrss

ml4gw/waveforms/adhoc/white_noise_burst.py
```diff
@@ -0,0 +1,217 @@
...
+        # f_phys: shape (1, nfreq)
+        f_phys = k * df
+        factor = (2 * torch.pi * f_phys) ** 2
+        factor *= 2  # Account for negitive frequency
+        power_plus = torch.sum(factor * (torch.abs(Hplus) ** 2), dim=-1) * df
+        power_cross = torch.sum(factor * (torch.abs(Hcross) ** 2), dim=-1) * df
+        current_hdotsq = power_plus + power_cross  # shape (batch,)
+        eps = 1e-50
+
+        norm_factor = torch.sqrt(
```
wbenoit26, line 197: Similar issue here. Normalization is computed across the full array, but the returned array is a slice, so you don't actually get the requested power.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+        Generate the chosen cosmic-string waveform in plus polarization,
+        with cross=0.
+        waveform must be:
+            "cusp"  -4.0 / 3.0,
+            "kink" -5.0 / 3.0, or
+            "kinkkink" -2.0.
+        Args:
+            power: cusp = -4.0 / 3.0, kink = -5.0 / 3.0, or kinkkink = -2.0.
+            amplitude: (batch,) overall amplitude scaling parameter.
+            f_high: (batch,) freq above which we apply exponential taper.
```
wbenoit26, line 89: `power` and `f_high` are listed as `(batch,)` here, but as floats in the signature.

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+        f_clamped = torch.clamp(freq, min=1e-20)
+
+        # Common factor => (1 + (f_low^2)/(f^2))^-4
+        base_factor = (1.0 + (self.f_low**2) / (f_clamped**2)) ** (-4.0)
+
+        # Multiply by f^power
+        base_factor *= f_clamped ** (power)
+
+        # Taper above f_high => if freq>f_high => exp(1 - freq/f_high), else 1
+        ratio = freq / f_high  # (batch, freq_bins)
```
wbenoit26, line 124: If `f_high` is of shape `(batch,)`, this line will raise an error

ml4gw/waveforms/adhoc/cosmic_string.py
```diff
@@ -0,0 +1,157 @@
...
+    ) -> (Tensor, Tensor):
+        """
+        Generate the chosen cosmic-string waveform in plus polarization,
+        with cross=0.
+        waveform must be:
+            "cusp"  -4.0 / 3.0,
+            "kink" -5.0 / 3.0, or
+            "kinkkink" -2.0.
+        Args:
+            power: cusp = -4.0 / 3.0, kink = -5.0 / 3.0, or kinkkink = -2.0.
```
wbenoit26, line 87: The options for `power` should be enforced, either with an enum or with a string selector.

wbenoit26 (changes requested):
@AndyC80297 I began to review, but I think there's actually a lot to say here. Could you split these into multiple PRs, one per waveform? That will make it easier to review and iterate.

PR #197, opened 2025-02-06
Fix flaky tests

EthanMarx:
- Fixes power law flaky test by removing initial guess
- Fixes spectral density flaky tests by allowing at most 1 sample to mismatch by more than tolerance. This happens occasionally, at it appears to be for low values. Eventually, the underlying cause of this should be addressed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,7 +27,7 @@ def compare_against_numpy():
     of the time
     """
 
-    def compare(value, expected):
+    def compare(value, expected, num_bad: int = 0):
         sigma = 0.01
         prob = 0.99999
         N = np.prod(expected.shape)
@@ -40,7 +40,7 @@ def compare_against_numpy():
 
         # TODO: eventually we should track down
         # and address the underlying cause
-        assert isclose.sum() - np.prod(isclose.shape) <= 1
+        assert isclose.sum() - np.prod(isclose.shape) <= num_bad
 
     return compare
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,10 +28,11 @@ def compare_against_numpy():
     """
 
     def compare(value, expected):
-        sigma = 0.01
-        prob = 0.9999
+        sigma = 0.02
+        prob = 0.99999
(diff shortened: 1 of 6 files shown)


EthanMarx:
@wbenoit26 The power law test was fixed by removing the initial guess. Spectral density test fixed by allowing at most 1 sample to differ by more than tolerance. This is a bandaid solution, but considering its only ever 1 sample and not that often, I think its acceptable. 

Ran the power law test 1000 and the spectral density 100 times with no errors

tests/conftest.py
```diff
@@ -17,10 +17,18 @@ def compare_against_numpy():
 
     def compare(value, expected):
         sigma = 0.01
-        prob = 0.9999
+        prob = 0.99999
```
wbenoit26, line 20: I was looking into this the other day - changing the probability has only a small effect on the tolerance, and wouldn't have prevented most of the failed unit tests (at least, for the ones that I spot checked). Changing the `sigma` value would have a larger impact; however, it also seemed like the relative error had a much smaller variance than the sigma we use here (around 0.001 for the fast spectral density with y), which makes me wonder if there's something genuinely wrong with the function for some inputs. Or if there's something wrong with the logic of the test.

wbenoit26: `test_fast_spectral_density_with_y` seems to be where it fails most often. Can you try running that one?

EthanMarx: yup running that now 100 times, close to completing

EthanMarx: That one ran but iirfilter tests are flaky.....

wbenoit26: True, and those tests do only one sample, so that doesn't bode well. There's some edge effects with this filter compared with SciPy, right? Maybe more of the edge needs to be sliced off.

EthanMarx:
@wbenoit26 The `iirfilter` is still flaky but don't want to hold these improvements up. I'll assign that to ravi

wbenoit26 (approved):
Sweet, this looks good to me

PR #207, opened 2025-03-06
Bug when calculating log prob of samples outside cosine distribution

EthanMarx:
Fixes bug where the `log_prob` of samples outside the cosine distribution was given a value of 0 instead of `-inf`


deepchatterjeeligo (approved):
Thanks for catching that

PR #210, opened 2025-04-24
Add data download to example notebook

wbenoit26:
Quick update to the example notebook so that users can download the data themselves rather than pointing them to a location on the LDG.


wbenoit26:
@EthanMarx Could we get this merged in?

PR #211, opened 2025-05-15
Benchmarking

sjhend03:
Added code and scripts to benchmark the cbc waveform generation. `benchmark.ipynb` explains the code.


sjhend03:
@wbenoit26 Thanks, I used the pre-commit hooks and fixed the conflict (I believe)

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -136,18 +247,83 @@ def test_taylor_f2(
...
+            )
+            assert np.allclose(
+                1e21 * hc_lal_data.imag,
+                1e21 * hc_ml4gw.imag.numpy(),
+                atol=1e-3,
+            )
+
+        # If benchmarking calculate errors to record
+        # test_2 refers to differences between lal and ml4gw
+        # TaylorF2 methods when mass ratio is flipped
```
wbenoit26, line 274: I suggest that we remove the benchmarking for the flipped mass ratio test. I think really we should be comparing `ml4gw` waveforms to each other for this test, not to lal, though that doesn't need to be in your PR

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -136,18 +247,83 @@ def test_taylor_f2(
...
+
+        # append data to batch and write if batch size exceeds set batch size
+        if request.config.getoption("--benchmark") != 0:
+            batch_data.append(data)
+            if len(batch_data) >= batch_size:
+                for batch in batch_data:
+                    write_benchmark_data(filename, batch)
+                batch_data = []
+
+                if get_file_size(filename) >= 0.05:
```
wbenoit26, line 318: If this check is here, you can remove the check in `get_next_file_name`

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -136,18 +247,83 @@ def test_taylor_f2(
...
+                    filename = get_next_file_name(base_filename)
+                    logging.info(f"Switching to new file: {filename}")
+
+    # If there is any remaining data in the batch, write it to the file
+    if request.config.getoption("--benchmark") != 0:
+        if batch_data:
+            for batch in batch_data:
+                write_benchmark_data(filename, batch)
```
wbenoit26, line 326: After writing everything to file, it would be great to have a function to aggregate all the files and datasets into a single file

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -160,11 +336,18 @@ def test_phenom_d(
...
     sample_rate,
     f_ref,
+    request,
 ):
     mass_1, mass_2 = chirp_mass_and_mass_ratio_to_components(
         chirp_mass, mass_ratio
     )
 
+    # Accumulate data in memory and write in batches for benchmarking only
+    batch_size = 100
```
wbenoit26, line 346: You've got `batch_size` as a fixture above

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -414,18 +690,87 @@ def test_phenom_p(
...
+                atol=2e-2,
+            )
+            assert np.allclose(
+                1e21 * hc_lal_data.imag,
+                1e21 * hc_ml4gw.imag.numpy(),
+                atol=2e-2,
+            )
+
+        # If benchmarking calculate errors to record
+        # test_2 refers to differences between lal and ml4gw
```
wbenoit26, line 716: The difference between test 1 and test 2 is that distances are greater than 400 Mpc for test 1 and less than 400 Mpc for test 2. I'd suggest renaming test 1 and test 2 as "far" and "close"

tests/waveforms/cbc/test_cbc_waveforms.py
```diff
@@ -446,38 +791,40 @@ def test_phenom_p(
...
         assert np.allclose(
-            1e21 * hc_ml4gw[0].imag.numpy(),
-            1e21 * hc_ml4gw[i + 1].imag.numpy(),
-            atol=1e-2,
+            1e21 * hc_lal_data.imag, 1e21 * hc_ml4gw[0].imag.numpy(), atol=2e-2
         )
+
+    # check batch against each other
+    for i in range(9):
+        if request.config.getoption("--benchmark") == 0:
```
wbenoit26, line 810: Nitpick, but this `if` statement can be outside the for-loop

plot_benchmark.py
```diff
@@ -0,0 +1,156 @@
+import os
```
wbenoit26, line 1: This file and `benchmark.ipynb` should be moved into an `examples/benchmarks` directory. And the current contents of the `examples` directory should be moved into an `examples/tutorials` directory

wbenoit26 (changes requested):
This looks great! I want to iterate a bit on the way that the benchmarking is integrated into the testing setup - I think there are some ways we can optimize. I haven't commented much on the notebook or the plotting code because I think those will need to change based on modifications to the testing.

wbenoit26:
We should circle back on this. None of this is currently in the documentation, but it would be great to get it in there. I have some changes requested above, and I'm seeing some changes to the whitener that should be in a different PR.

PR #212, opened 2025-05-26
Fix cache when using external data directories (needed it for gwak)

asasli:


wbenoit26:
@asasli Can this be closed?

PR #213, opened 2025-05-27
Improve test coverage

wbenoit26:
I added some tests to cover statements we were missing for some of the more commonly used functions. Still plenty of places that need coverage, but this makes some progress.


wbenoit26:
@deepchatterjeeligo Could you take a look at this when you get a chance, particularly the changes to IMRPhenomD? I'm fairly confident that the checks on `fRD` and `fDM` were unnecessary (and all the tests are passing), but it would be good to get another pair of eyes.

deepchatterjeeligo (approved):
At this point I don't recall the reason for placing the conditions on `fRD` and `fDM`. I am happy to make this change. The  fact that the tests pass without them gives me some relief. We can revert it in case we run into problems.

PR #216, opened 2025-06-30
Optionally return indices when sampling kernels

wbenoit26:
Adds an option to return `idx` from `sample_kernels` so that it can be known where each sample ended up.


deepchatterjeeligo (approved):
This is backward compatible, and looks good. Feel free to merge whenever @wbenoit26

PR #222, opened 2025-08-22
drop support for python3.9 (EOL in Oct 2025)

deepchatterjeeligo:
- update workflows
- ensure PEP compliant license field

### DIFF
--- a/ml4gw/dataloading/in_memory_dataset.py
+++ b/ml4gw/dataloading/in_memory_dataset.py
@@ -191,7 +191,9 @@ class InMemoryDataset(torch.utils.data.IterableDataset):
             # indices we'll need rather than having to generate
             # everything.
             idx = [range(self.num_kernels) for _ in range(len(self.X))]
-            idx = zip(range(num_kernels), itertools.product(*idx), strict=True)
+            idx = zip(
+                range(num_kernels), itertools.product(*idx), strict=False
+            )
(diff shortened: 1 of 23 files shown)


wbenoit26:
What changes with the removal of 3.9 that we need to ignore the zip check?

EthanMarx:
Yeah, that [check](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) ensures that we are using the `strict` parameter explicitly, so I think we need to keep it and add the `strict` argument where we perviously weren't using it

deepchatterjeeligo:
OK I applied the strict change. The diff is fairly large but straightforward.

deepchatterjeeligo:
anything else?

deepchatterjeeligo:
Yeah strict=True does not work everywhere. I'll look at it later.

ml4gw/dataloading/in_memory_dataset.py
```diff
@@ -191,7 +191,9 @@ def init_indices(self):
             # indices we'll need rather than having to generate
             # everything.
             idx = [range(self.num_kernels) for _ in range(len(self.X))]
-            idx = zip(range(num_kernels), itertools.product(*idx))
+            idx = zip(
+                range(num_kernels), itertools.product(*idx), strict=False
```
deepchatterjeeligo, line 195: Turns out this the only place where we need the strict=False for zip since the itertools.product can return a longer iterator

PR #226, opened 2025-09-24
Add decimator transform for downsampling timeseries data

bhgupta20:
Implements a decimation transform to reduce the sampling rate of the input time series based on a schedule. Also includes unit test in `test_decimator.py`.

### DIFF
--- a/ml4gw/transforms/decimator.py
+++ b/ml4gw/transforms/decimator.py
@@ -8,7 +8,7 @@ class Decimator(torch.nn.Module):
     The schedule is a tensor of shape (N, 3), where each row is:
     [start_time, end_time, target_sample_rate]
 
-    Example default schedule (seconds, Hz):
+    Example schedule (seconds, Hz):
     [[0, 40, 256],   # first 40s at 256 Hz
      [40, 58, 512],  # next 18s at 512 Hz
      [58, 60, 2048]] # last 2s at full 2048 Hz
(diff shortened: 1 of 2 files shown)


ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,195 @@
...
+        device=None,
+    ):
+        super().__init__()
+        self.sample_rate = sample_rate
+        self.schedule = schedule
+        self.device = device
+
+        if sample_rate is not None and schedule is not None:
+            self._validate_inputs(sample_rate, schedule)
+            self.idx = self.build_variable_indices(
```
wbenoit26, line 39: When you calculate `idx` here, you can store it as a buffer with `register_buffer`; there are some examples in the other transforms. This will allow you to not need to specify the device anywhere.

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,195 @@
...
+        self,
+        sample_rate: int,
+        schedule: torch.Tensor,
+        strain: torch.Tensor = None,
+    ):
+        """
+        Validate inputs for compatibility.
+        If `strain` is None, only validates schedule and sample_rate.
+        """
+        if not isinstance(sample_rate, int) or sample_rate <= 0:
```
wbenoit26, line 55: Testing for type and for positivity is probably overkill. There's nothing wrong with doing so, but it's reasonable to expect the user to do something sensible here. Also, in your testing, you should make sure that these errors actually get raised. You can see the process for doing so in some of our other tests, and you can see the testing coverage report in the comments on this PR; let me know if you have any questions about it.

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,195 @@
...
+        schedule: torch.Tensor = None,
+        device=None,
+    ):
+        """
+        Build a mask of indices to decimate a strain according to the schedule.
+        """
+        sample_rate = (
+            sample_rate if sample_rate is not None else self.sample_rate
+        )
+        schedule = schedule if schedule is not None else self.schedule
```
wbenoit26, line 106: I think these checks are redundant with the checks in `forward`

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,195 @@
...
+
+        self._validate_inputs(sample_rate, schedule, strain=strain)
+
+        if (
+            self.idx is not None
+            and sample_rate == self.sample_rate
```
wbenoit26, line 180: This won't work if a user passes a sample rate initially, but not during the forward call. I would suggest having the pattern be that the sample rate and schedule _must_ be passed on initialization, and then can't be passed during `forward`. If a user wants to do a different decimation, they should create a new `Decimator` object.

wbenoit26 (changes requested):

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,137 @@
...
+                    f"target rates {self.schedule[:, 2].tolist()}"
+                )
+
+        else:
+            if strain.ndim < 1:
+                raise ValueError(
+                    "strain must have at least 1 dimension (time axis)"
+                )
+
+            strain_len = int(
```
wbenoit26, line 72: This is something that could be calculated during `__init__`, and then in `forward` we could just check if `self.expected_size == strain.shape[-1]`. And then this validation function could be simplified so that it never needs to take a `strain` argument (I think it's fine to remove the check for strain dimensionality).

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,137 @@
...
+
+    The schedule is a tensor of shape (N, 3), where each row is:
+    [start_time, end_time, target_sample_rate]
+
+    Example schedule (seconds, Hz):
+    [[0, 40, 256],   # first 40s at 256 Hz
+     [40, 58, 512],  # next 18s at 512 Hz
+     [58, 60, 2048]] # last 2s at full 2048 Hz
+
+    Strain: Tensor of shape (B, C, T), where
```
wbenoit26, line 16: I should've thought of this earlier: this transform is generically applicable to timeseries, so we should keep the variable names and documentation generic too. Could you go through and update the terminology? It's fine to use something like `X` for the data name, and just say something like "timeseries" instead of strain or waveforms.

ml4gw/transforms/decimator.py
```diff
@@ -0,0 +1,137 @@
+import torch
+
+
+class Decimator(torch.nn.Module):
+    """
+    Downsample/decimate timeseries according to a schedule.
+
+    The schedule is a tensor of shape (N, 3), where each row is:
+    [start_time, end_time, target_sample_rate]
+
```
deepchatterjeeligo, line 10: add in the docstring that this is a naive decimator that is not as sophisticated in the sense it does not use any IIR/FIR filters.

wbenoit26 (approved):

PR #237, opened 2025-10-22
Add `crop` parameter to whitening

wbenoit26:
This PR adds a `crop` parameter to the `whiten` function and the whitening transforms that allows data to be returned uncropped. By default, `crop = True`.


deepchatterjeeligo (approved):
Looks good.

PR #239, opened 2025-11-03
adding the overlap functionality to the schedule in `decimator` transform

bhgupta20:
made changes to the `decimator` transform by adding overlapping windows to the schedule


wbenoit26 (approved):
Looks great, @bhgupta20. 

Unrelated to this PR, I think it would be beneficial to add a utility function to this class that returned the times associated with each sample in each segment.

PR #240, opened 2025-11-18
 Implement PhenomDECO - a BBH waveform to infer compactness 

Shrobana:
PhenomDECO extends the aligned-spin binary black hole waveform PhenomD by
modifying the amplitude; the phase model is unchanged. The modification is
parameterized by the effective compactness, enabling direct inference of compactness from
data.

Changes include 

- Add new PhenomDECO CBC waveform class
- Add new unit test for comparing PhenomDECO to PhenomD at the BBH limit
- Add example notebook for generating the waveform


ml4gw/waveforms/cbc/phenom_deco.py
```diff
@@ -0,0 +1,359 @@
...
+        chi2: BatchTensor,
+        compactness: BatchTensor,
+        distance: BatchTensor,
+        phic: BatchTensor,
+        f_ref: float,
+    ) -> Float[FrequencySeries1d, " batch"]:
+        # PhenomDECO reuses the PhenomD phase model and the amplitude model
+        # is parametrized by an effective compactness of the two objects;
+        # see Phys. Rev. D 112, 104017 (2025) for details
+
```
deepchatterjeeligo: could you convert this into a docstring? It will render in the documentation page automatically.

ml4gw/waveforms/cbc/phenom_deco.py
```diff
@@ -0,0 +1,359 @@
...
+            phic,
+            f_ref,
+        )
+
+        hp = (htilde.mT * pfac).mT
+        hc = -1j * (htilde.mT * cfac).mT
+
+        return hc, hp
+
+    def phenom_d_htilde(
```
deepchatterjeeligo: would it better to call this `phenom_deco_htilde`? Since the phase is the same as phenom_d but the amplitude gets modified w.r.t. phenom_d.

deepchatterjeeligo (changes requested):
Hi @Shrobana thank you for this nice atomic contribution. This is ready in terms of the content. I left some comment w.r.t. cosmetics and documentation.

examples/phenom_deco.ipynb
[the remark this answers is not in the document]
wbenoit26: This could also go in the [examples](https://github.com/ML4GW/ml4gw/tree/main/docs/examples) section of the documentation, as it looks relatively short. But either way, it'll be good to have.

ml4gw/waveforms/cbc/phenom_deco.py
Shrobana: Renamed this function

examples/phenom_deco.ipynb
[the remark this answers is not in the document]
Shrobana: Added a more informative docstring in the code and brief documentation with the example [here](https://github.com/Shrobana/ml4gw/blob/dev/docs/tutorials/ml4gw_tutorial.ipynb).

Issue #142, opened 2024-06-27
Constrain `numpy<2.0.0`


EthanMarx:
Something changed in the numpy api, causing our `Hdf5TimeSeriesDataset` to not properly fill the returned array.

PR #241, opened 2025-11-25
Upgrade `numpy` to >2

benedict-armstrong:
https://github.com/ML4GW/ml4gw/issues/142 requested the `numpy` version to be <2 due to a change in the `np.unique` (see: https://numpy.org/doc/stable/reference/generated/numpy.unique.html version 2.0 changes). The proposed solution (suggested in the docs) is to flatten `inv` returned by `np.unique`.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -35,3 +35,37 @@ abstract: >-
 keywords:
   - Gravitational waves
   - Machine learning
+preferred-citation:
+  type: article
+  authors:
+  - given-names: William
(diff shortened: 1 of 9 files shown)


benedict-armstrong:
Hi @wbenoit26 thanks for responding so quickly. Good point, I've changed the PR to just remove the <2 restriction. I've tested both `numpy`<2 and `numpy`>2 locally. Please excuse the messy git history, this is my first real PR 😅

PR #261, opened 2026-04-07
Update python versions

wbenoit26:
Just seeing what tests fail, if any, when changing the python versions to >=3.11,<3.15.


deepchatterjeeligo (commented):
This is good to go once PR is against `dev`.

deepchatterjeeligo (approved):

PR #262, opened 2026-04-08
Create dependabot.yml

wbenoit26:


.github/dependabot.yml
deepchatterjeeligo, line 1: Seems like this will propose updates to pyproject.toml, but we will need to sync our uv.lock file after? BTW that's fine with me. We can simply push the new lock file to the PR that the bot creates.

deepchatterjeeligo: But that should not hold this PR back. Hence I will approve.

deepchatterjeeligo (approved):

.github/dependabot.yml
wbenoit26: Good point, I hadn't considered that. It [looks like](https://docs.astral.sh/uv/guides/integration/dependabot/) using `package-ecosystem: "uv"` will also update the lock files; I'll make that change.

PR #264, opened 2026-04-09
Update maintenance tools

wbenoit26:
- Switch from pre-commit to [prek](https://prek.j178.dev), which is a drop-in replacement and seems like a good tool moving forward.
- Move to a more standard unit test coverage reporting tool, [Codecov](https://about.codecov.io)
- Use PyPI's trusted publisher feature to enable GitHub actions to publish releases without needing to store the token in the repo.


deepchatterjeeligo (changes requested):
Looks good. Could you change the instructions in the contribution guide to use prek instead of pre-commit?

wbenoit26:
Good catch, I've made that change.

deepchatterjeeligo (approved):

PR #266, opened 2026-04-10
Remove tox

wbenoit26:
We're not doing anything with `tox` that can't be straightforwardly handled with the testing matrix, so I've removed it from the workflow to simplify things. I've also moved the dependencies that appear only in the tutorial notesbooks into the `docs` group.


deepchatterjeeligo (approved):
Looks good

PR #278, opened 2026-04-13
Use OIDC for codecov

wbenoit26:
CI fails for PRs created by dependabot because it doesn't have token access. This PR replaces the token-based authentication with OIDC.


wbenoit26:
@deepchatterjeeligo Just going bypass review for this one so that I can clean up the dependabot PRs more quickly. If it causes issues I'll sort it out.

PR #287, opened 2026-04-17
Use asd instead of psd

### LINKED ISSUE #70: ASD vs. PSD
Right now the default representation of spectral density that is expected by most functions is the PSD. However, most of these functions end up converting to the ASD internally, and the ASD has much more desirable numerical properties since the PSD for strain is typically O(10^-40-10^-45), which is getting dangerously close to 0 for 32 bit precision. Given this, does it make sense to have any all spectral density calculators return the ASD and have any functions that expect a spectral representation expect the ASD? Which functions would have to convert back, and how much of a performance impact would this have? (presumably pretty small)

wbenoit26:
Closes #70

### DIFF
--- a/ml4gw/gw.py
+++ b/ml4gw/gw.py
@@ -15,10 +15,10 @@ from torch import Tensor
 
 from .constants import C
 from .types import (
+    ASDTensor,
     BatchTensor,
     NetworkDetectorTensors,
     NetworkVertices,
-    PSDTensor,
     TensorGeometry,
     VectorGeometry,
     WaveformTensor,
@@ -283,7 +283,7 @@ def get_ifo_geometry(
 
 def compute_ifo_snr(
     responses: WaveformTensor,
-    psd: PSDTensor,
+    asd: ASDTensor,
     sample_rate: float,
     highpass: float | Float[Tensor, " frequency"] | None = None,
     lowpass: float | Float[Tensor, " frequency"] | None = None,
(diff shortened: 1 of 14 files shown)


deepchatterjeeligo (commented):
This change of change will likely break (or may not, but give unexpected result) for downstream codes. What is motivation for this change? We should set a milestone for these kinds of changes.

wbenoit26:
This has been a long-standing to-do item (you can see that Alec created the issue more than 2.5 years ago), and in fact it was part of the [v0.5.0 milestone](https://github.com/ML4GW/ml4gw/milestone/1), which clearly didn't happen. The main motivation is to avoid the kind of silent failures that occur when PSD values get zeroed out when users pass float32 data. This could also be addressed by raising an error if any of the values are zero, but I think the right approach is this one. As a side benefit, there should be some minor performance benefits from the reduced precision.

Incorporating this change will certainly break things in downstream projects. I was planning to make the next release be v0.9.0 and simultaneously restrict ml4gw versions in our projects if we don't want to make the relevant modifications right away.

deepchatterjeeligo:
I am OK with making a future proofing change as long as it is not disruptive to more immediate plans: I am thinking IR1. If this is material for 0.9.0 milestone, and 0.9.0 has any possibility of being included in our stack for IR1, then this is invasive atm. All this is to say that this should be a post IR1 milestone.

wbenoit26:
Yeah, I agree. I'm happy to wait until we settle on final versions for our IR1 algorithms before merging this in.

deepchatterjeeligo (commented):
@wbenoit26 once you resolve the merge conflicts, can you make this PR against `dev`?

deepchatterjeeligo (approved):
Looks good!

PR #289, opened 2026-04-20
Improve waveform testing

wbenoit26:
Replaces randomly sampling a large number of waveforms with a small set of specific test cases. The testing condition is also changed from element-wise comparison to an overlap measurement. For the waveform generator, a further tapering function is done, in line with what gwsignal does.

@deepchatterjeeligo Curious to get your thoughts on this approach.


deepchatterjeeligo (approved):
Looks good. Can I still suggest you make a quick plot of the residuals? I am curious where were the residuals the highest. I suspect around the merger. Would be great to see it in a figure.

But I'll still approve the PR.

wbenoit26:
[Here](https://ldas-jobs.ligo-wa.caltech.edu/~william.benoit/cbc_residuals/) are plots of all the residuals for each test. The TD residuals are relatively large I think because of the need to align the ml4gw and gwsignal waveforms to the nearest sample.

PR #299, opened 2026-05-19
Bugfixes

wbenoit26:
Fixes a few miscellaneous bugs:

1. In `get_data_indices`: clamp to `[0, N//2]` to prevent `IndexError` when `frange[1]` is near the Nyquist frequency
2. In `get_initial_state`: `device=self.weights.device` so state tracks the model's device
3. In `test_decimator.py`: `param_dict = {k: v.to(device) ...}` before calling `waveform_generator` prevents tests from failing when running on a node with a GPU.
4. Restricts the torch versions to <2.11. Starting with 2.11, torch builds against CUDA 13 by default, which is incompatible with V100s.


bhgupta20 (approved):
This looks good to me!

PR #301, opened 2026-05-23
Benchmarking

wbenoit26:
Creates a benchmarking suite for most of the public modules/functions. Also adds instructions in for running benchmarks and performing comparisons.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,6 +76,18 @@ if an automatic fix is not possible.
    usecase.
 4. Please [cite](/CITATION.cff) the repository if you have used `ml4gw` in your work.
 
+## Benchmarks
+
+`ml4gw` maintains a performance benchmark suite under [`benchmarks/`](benchmarks/) using
+[pytest-benchmark](https://pytest-benchmark.readthedocs.io/). See the
+[benchmarks README](benchmarks/README.md) for full instructions on running the suite and
+measuring the impact of a change.
+
+When submitting a PR that modifies a performance-sensitive code path (e.g. a transform,
+waveform model, or spectral function), **run the relevant benchmarks on GPU hardware** and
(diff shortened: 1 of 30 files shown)


CONTRIBUTING.md
```diff
@@ -76,6 +76,18 @@ if an automatic fix is not possible.
...
+
+`ml4gw` maintains a performance benchmark suite under [`benchmarks/`](benchmarks/) using
+[pytest-benchmark](https://pytest-benchmark.readthedocs.io/). See the
+[benchmarks README](benchmarks/README.md) for full instructions on running the suite and
+measuring the impact of a change.
+
+When submitting a PR that modifies a performance-sensitive code path (e.g. a transform,
+waveform model, or spectral function), **run the relevant benchmarks on GPU hardware** and
+paste the `--benchmark-compare` table into your PR description so reviewers can see the
+before/after delta.
```
deepchatterjeeligo, line 89: Good to also add what should be the route when adding a new transform or waveform as opposed to modifications. For example, if there should be unittesting code along with new benchmark code.

deepchatterjeeligo (commented):
Great addition @wbenoit26. I have never used pytest-benchmark plugin but seems nifty. I also like the new policy of providing the comparison when touching performance sensitive features. I have a minor comment about documentation. Also wondering if you have found whether the unittests can be combined with benchmarks, since some of the test are probably repeats. I am not particular about this, however. Keeping the unittests and benchmarks separate is also a fair point.

CONTRIBUTING.md
wbenoit26: Good idea, I'll add that.

wbenoit26:
I found it mentally easier to keep the unit tests and benchmarks separate. The benchmarks are all pretty simple, at least for now; we may end up wanting to benchmark across more parameters than just batch size).

deepchatterjeeligo (approved):

PR #304, opened 2026-05-28
Chirp distance and mass joint distribution

aggar104:
Jointly sample chirp mass and luminosity distance according to the chirp distance scaling.


ml4gw/distributions.py
```diff
@@ -391,3 +391,66 @@ def _p_of_distance(self):
         # This is a tensor of ones if the distance type is redshift
         jacobian = torch.gradient(self.distance_grid, spacing=self.dz)[0]
         return dV_dz / jacobian * self.rate_function(self.z_grid)
+
+
+class ChirpDistribution(dist.Distribution):
+    def __init__(
+        self,
+        chirp_mass: torch.distributions.Distribution,
+        chirp_distance: torch.distributions.Distribution,
```
deepchatterjeeligo, line 400: Why do we need a chirp_distance as an input distribution?

ml4gw/distributions.py
```diff
@@ -391,3 +391,66 @@ def _p_of_distance(self):
...
+
+        return mass_factor
+
+    def sample(
+        self,
+        N: torch.Size,
+    ):
+
+        chirp_mass = self.chirp_mass.sample(N)
+        chirp_distance = self.chirp_distance.sample(N)
```
deepchatterjeeligo, line 428: So I was hoping that we call this distribution `UniformChirpDistance` meaning the result chirp distance distribution is uniform.

deepchatterjeeligo (changes requested):
Hi @aggar104 thanks for starting this. I think the overall idea is correct. I was hoping for this to be a little simpler and be Uniform in chirp distance. Could you also share a plot which should what the samples look like? And also please add some nice unittests.

deepchatterjeeligo:
@aggar104 gentle bump on this.

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+        self,
+        chirp_mass: torch.distributions.Distribution,
+        chirp_distance: torch.distributions.Distribution,
+        reference_chirp_mass: float = 1.0,
+    ):
+        super().__init__()
+
+        self.chirp_mass = chirp_mass
+        self.chirp_distance = chirp_distance
+        self.ref_mass = ref_mass
```
wbenoit26, line 407: The input name is `reference_chirp_mass`

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+    ):
+        """
+        Evaluate prior in transformed coordinates.
+        """
+
+        mass_factor = self.scale(chirp_mass)
+        chirp_distance = distance / mass_factor
+
+        # log_probs
+        logp_mass = self.chirp_mass.log_prob(chirp_mass)
```
wbenoit26, line 538: `self.chirp_mass` doesn't exist in this class.

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+        self,
+        chirp_distance: torch.distributions.Distribution,
+        reference_chirp_mass: float = 1.0,
+    ):
+        super().__init__()
+
+        self.chirp_distance = chirp_distance
+        self.reference_chirp_mass = reference_chirp_mass
+
+        self.joint_parameters = ["chirp_mass", "chirp_distance"]
```
wbenoit26, line 470: This doesn't get used anywhere.

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+        log_jacobian = -torch.log(mass_factor)
+
+        return logp_mass + logp_distance + log_jacobian
+
+
+class ChirpDistanceUniform(dist.Distribution):
+    def __init__(
+        self,
+        chirp_distance: torch.distributions.Distribution,
+        reference_chirp_mass: float = 1.0,
```
wbenoit26, line 463: We should default to a 1.4+1.4 system, so ~1.22.

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+
+        # log_probs
+        logp_mass = self.chirp_mass.log_prob(chirp_mass)
+        logp_distance = self.chirp_distance.log_prob(chirp_distance)
+        log_jacobian = -torch.log(mass_factor)
+
+        return logp_mass + logp_distance + log_jacobian
+
+
+class ChirpDistanceUniform(dist.Distribution):
```
wbenoit26, line 459: Let's call this `UniformChirpDistance`

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+    def __init__(
+        self,
+        chirp_distance: torch.distributions.Distribution,
```
wbenoit26, line 462: Given that this is supposed to be a uniform chirp distance, we shouldn't need to take `chirp_distance`.

ml4gw/distributions.py
```diff
@@ -391,3 +391,152 @@ def _p_of_distance(self):
...
+        chirp_mass: torch.Tensor,
+        chirp_distance: torch.Tensor,
+        **kwrags,
+    ):
+
+        mass_factor = self.scale(chirp_mass)
+        distance = chirp_distance * mass_factor
+        return distance
+
+    def sample(
```
wbenoit26, line 496: I think we shouldn't have the `sample` method return a dictionary, as that's different from what a `Distribution` does.

wbenoit26 (changes requested):
I think we can make this a lot simpler, and it should just be a single class. I've left some comments on the code, but I think a full re-write is the correct approach. If I understand, the goal is to have a class that returns a tensor of `luminosity_distance` values that are uniform in chirp distance, given a tensor of chirp masses.

Because `d_L = d_c (M_c / M_ref) ** (5/6)`, a uniform chirp distance implies that `d_L` follows `Uniform(s * minimum, s * maximum)` for each `M_c` value, where `s` is the scale factor. While looking into this, I learned that torch's `Uniform` distribution can take a tensor of minimum and maximum values, and when you sample from that distribution, you get one sample per (min, max) pair. So what we could do is something like:

```python
class UniformChirpDistance:
    def __init__(self, ...):
        ...

    def __call__(self, chirp_mass):
        scale = (chirp_mass / reference_chirp_mass) ** (5 / 6)
        return Uniform(
            minimum * scale,
            maximum * scale,
        )
```

(With proper docstrings and type annotation, of course). And then sampling from the returned `Uniform` distribution would give you your luminosity distance.

PR #305, opened 2026-05-29
Add benchmark comparison script

wbenoit26:
Adds a convenience script for more easily viewing the difference between two benchmarks. Updates the instructions for using this script.


deepchatterjeeligo (approved):
Nice addition.

PR #311, opened 2026-06-03
Add s4d

kyoon-mit:
## Summary

Adds the S4D state-space model architecture to `ml4gw`.

- New model in `ml4gw/nn/ssm/s4d.py`.
- Tests in `tests/nn/ssm/test_s4d.py`
- Benchmark in `benchmarks/nn/test_s4d.py`: forward pass across batch size and testing O(L) behavior for sequence length L.

## Benchmark

Forward-pass medians on a shared NVIDIA A100.

| batch | L=512 | L=2048 | L=4096 |
|------:|------:|-------:|-------:|
| 32 | 4.0 ms | 12.7 ms | 23.7 ms |
| 128 | 10.9 ms | 41.9 ms | 84.5 ms |
| 512 | 40.4 ms | 162.2 ms | 328.8 ms |

- Confirms O(L) scaling.


wbenoit26 (changes requested):
Overall, this looks good to me. I'm not familiar with the details of what this architecture should look like, so it would be good to get a second pair of eyes from someone who is; maybe @chreissel or @benedict-armstrong could confirm.

Could you also add a couple small tests for the cases of `transpose=False`, `lr=0.0`, `lr=None`, and a test to make sure the dropout probability error gets raised cleanly?

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
+# Adapted with modifications from https://github.com/state-spaces/s4
+# (models/s4/s4d.py and src/models/nn/dropout.py), licensed under Apache-2.0.
+
+import math
+
+import torch
+import torch.nn as nn
+
+
+class DropoutNd(nn.Module):
```
deepchatterjeeligo: This seems general purpose, let's move it to a `dropout.py` similar to the `norm.py` under [nn](https://github.com/ML4GW/ml4gw/tree/main/ml4gw/nn)

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+    """N-dimensional dropout that ties the mask across sequence positions."""
```
deepchatterjeeligo: I don't understand the docstring. Please provide an example usage.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+            if not self.transposed:
+                X = X.movedim(-1, 1)
+            mask_shape = (
+                X.shape[:2] + (1,) * (X.ndim - 2) if self.tie else X.shape
+            )
+            mask = torch.rand(*mask_shape, device=X.device) < 1.0 - self.p
+            X = X * mask * (1.0 / (1 - self.p))
+            if not self.transposed:
+                X = X.movedim(1, -1)
+        return X
```
deepchatterjeeligo: This can be made more general purpose by avoiding `if self.training`. The assumption is whatever code calls this dropout block will only use it in the training loop.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class S4DKernel(nn.Module):
+    """Generate convolution kernel from diagonal SSM parameters."""
```
deepchatterjeeligo: Improve docstring. Describe the arguments. Also a highlight of what is being done.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+        self,
+        d_model: int,
+        N: int = 64,
+        dt_min: float = 0.001,
+        dt_max: float = 0.1,
+        lr: float | None = None,
+    ):
+        super().__init__()
+
+        H = d_model
```
deepchatterjeeligo: why not use `H` in the function signature

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+
+        C = C * (torch.exp(dtA) - 1.0) / A
+        K = 2 * torch.einsum("hn, hnl -> hl", C, torch.exp(K)).real  # (H, L)
+        return K
+
+    def register(
+        self, name: str, tensor: torch.Tensor, lr: float | None = None
+    ) -> None:
+        """Register a tensor with a configurable LR and 0 weight decay."""
+        if lr == 0.0:
```
deepchatterjeeligo: When is lr zero? Or in other words, can we design this so that we choose the "training" phase as the default. For the testing phase, the attribute can be changed in the test step.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,223 @@
...
+        log_dt = torch.rand(H) * (
+            math.log(dt_max) - math.log(dt_min)
+        ) + math.log(dt_min)
+
+        C = torch.randn(H, N // 2, dtype=torch.cfloat)
+        self.C = nn.Parameter(torch.view_as_real(C))
+        self.register("log_dt", log_dt, lr)
+
+        log_A_real = torch.log(0.5 * torch.ones(H, N // 2))
+        A_imag = math.pi * torch.arange(N // 2).repeat(H, 1)
```
deepchatterjeeligo, line 79:
```suggestion
        A_imag = torch.pi * torch.arange(N // 2).repeat(H, 1)
```
removes the `math` import

deepchatterjeeligo (changes requested):
@kyoon-mit Overall fine. I think however this design intermingles training vs. inference modes together. That role should be kept for downstream, like the train dataloader or the test dataloader. One thing that you can do is to choose one mode as a default, like the training mode i.e. keep `register_parameter` and zero weight decay by default. Also if you can clarify what `lr=0` does.

benchmarks/nn/test_s4d.py
```diff
@@ -0,0 +1,23 @@
+"""Benchmarks for the S4D sequence model."""
+
+import pytest
+import torch
+from constants import NUM_CHANNELS
+
+from ml4gw.nn.ssm.s4d import S4Model
+
+
+@pytest.fixture(params=[512, 2048, 8192], ids=lambda n: f"L_{n}")
```
deepchatterjeeligo: Just so that we have a number for `8192` than OOM, should be change this to `4096`

ml4gw/nn/ssm/s4d.py
deepchatterjeeligo: actually I realize that there are other places where `math` is used. So this can be ignored.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Fixed and committed.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Replaced custom dropout function with `torch.nn.Dropout1d`

ml4gw/nn/ssm/s4d.py
kyoon-mit: Replaced custom dropout function with `torch.nn.Dropout1d`.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Done.

ml4gw/nn/ssm/s4d.py
kyoon-mit: I think this was an intentional design choice by the authors. Internally, `H` (hidden dimension) gets passed around for consistency with the paper (2206.11893), but externally, `d_model` is used for clarity.

ml4gw/nn/ssm/s4d.py
kyoon-mit: It seems like the S4 authors left it there to experiment with which parameters can be freezed during training time (see Table 3. (b) of 2206.11893, for example). Since I don't think this is needed anymore, I removed it.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class S4DKernel(nn.Module):
+    """Build the convolution kernel for one S4D layer.
+
+    Each channel is an independent diagonal linear time-invariant
+    state-space model, defined in continuous time by
+
+        x'(t) = A * x(t) + B * u(t)
+        y(t)  = C * x(t)
```
deepchatterjeeligo: Use mathjax. See [these lines](https://github.com/ML4GW/ml4gw/blob/main/ml4gw/gw.py#L287-L292) for an example.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+    with A diagonal. Linear time invariance means the output equals the
+    input convolved with a single kernel K, so the whole sequence is
+    processed in one convolution instead of stepped through position by
+    position. Discretizing with timestep dt gives the kernel this module
+    returns:
+
+        K_l = 2 * Re(
+            sum_n C_n * ((exp(dt * A_n) - 1) / A_n) * exp(dt * A_n)^l
+        ),
+        l = 0, ..., L-1   (n runs over the N/2 states)
```
deepchatterjeeligo: Can we convert this to latex for an easier read. Also can you built the docs, and paste a screenshot of the rendered html. There are instructions in the contribution guide.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+            i.e. independent SSMs created.
+        N: State size per model dimension.
+            The states are stored as N / 2 conjugate pairs, so N must be even.
+        dt_min: Lower bound of the per-channel timestep, sampled
+            log-uniformly in [dt_min, dt_max]. The timestep sets the
+            timescale each SSM resolves.
+        dt_max: Upper bound of the per-channel timestep.
+        lr: Optional learning rate for `dt` and `A` parameters.
+            If None, these parameters are optimized with the same LR as the
+            rest of the model. If 0.0, these parameters are frozen.
```
deepchatterjeeligo: Let's get rid of this. If zero is a proxy for freezing parameters, we can handle that downstream.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+    Args:
+        d_model: Model dimension.
+            Equivalent to the number of channels in CNN nomenclature,
+            i.e. independent SSMs created.
+        N: State size per model dimension.
+            The states are stored as N / 2 conjugate pairs, so N must be even.
+        dt_min: Lower bound of the per-channel timestep, sampled
+            log-uniformly in [dt_min, dt_max]. The timestep sets the
+            timescale each SSM resolves.
+        dt_max: Upper bound of the per-channel timestep.
```
deepchatterjeeligo, line 50: thanks for adding this

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+
+        H = d_model  # Number of independent SSMs (one per channel).
+
+        ### SSM parameter initialization ###
+        # For each channel, draw a random timestep `dt`
+        # from a log-uniform distribution.
+        log_dt = torch.rand(H) * (
+            math.log(dt_max) - math.log(dt_min)
+        ) + math.log(dt_min)
+        self.register("log_dt", log_dt, lr)
```
deepchatterjeeligo: Sorry if I was unclear last time. What I meant is that let's get rid of `.register` method altogether. Instead this line will be

```suggestion
        self.register_parameter("log_dt", log_dt)
```
Setting the weight decay to zero for `log_dt` will be done as a part of the `configure_optimizer` method of a lightningmodule. See [this function](https://github.com/state-spaces/s4/blob/main/example.py#L247-L291) in s4d repo, and in particular [these lines](https://github.com/state-spaces/s4/blob/main/example.py#L268-L276) where the `_optim` for certain parameters are handled.

ml4gw/nn/ssm/s4d.py
```diff
@@ -0,0 +1,233 @@
...
+            y = y.transpose(-1, -2)
+        return y, None
+
+
+class S4Model(nn.Module):
+    """Full S4D sequence model for regression / classification.
+
+    Input:  (B, d_input, L)  — channels-first (aframe convention).
+    Output: (B, d_output)
+    """
```
deepchatterjeeligo, line 203: If this is the final usable model, let's add an example in the docstring for the usage.

deepchatterjeeligo (changes requested):
Hi @kyoon-mit I think this is good progress. But maybe I was not clear about the `.register` last time. Could you take another pass. There are some minor comments about the docs as well.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Fixed in bc5c88e836ed97e725205808e2941ebc67e832d3.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Fixed in bc5c88e836ed97e725205808e2941ebc67e832d3.

benchmarks/nn/test_s4d.py
kyoon-mit: Fixed in 61ed6560874ac2ed83b71a207c6c0c20d8248b23. Also, the table is updated in the initial PR comment.

ml4gw/nn/ssm/s4d.py
kyoon-mit: image

kyoon-mit: Please see c0720658287af999514632a4b2bd07750036ff0a for updates to docstrings.

ml4gw/nn/ssm/s4d.py
kyoon-mit: Example added in c0720658287af999514632a4b2bd07750036ff0a.

kyoon-mit:
@wbenoit26 The optional `lr` parameter is now gone. The suggested dropout guard and the `transpose=False` test are added in bd8b67728a0d924dcd46311326ba2f9845e3c2d6.

Final version in c1403f7b359f135973cc8f4967b562bcde5a757e (removed meaningless None returns from the forward method).

deepchatterjeeligo (approved):
LGTM

wbenoit26 (approved):

PR #314, opened 2026-06-03
Optimize transforms

wbenoit26:
Made a number of optimizations to some of the slower transforms. @deepchatterjeeligo The scope of this PR expanded a bit beyond my original intention, so I'm happy to split this into multiple PRs if you'd prefer to review it that way.

The only real change to the public-facing API is that the return value of `unfold_windows` is now a view, rather than a copy. This means that some care needs to be taken with in-place operations. The benefit is that there's no intermediate 4-D tensor created. I've added a note to the function and confirmed that this doesn't affect anything in Aframe.

### Benchmarks (A100)

| Benchmark                                                     |  Baseline |   Current | Delta             |
|---------------------------------------------------------------|-----------|-----------|-------------------|
| test_spline_1d_forward[batch_32-num_points_128]               | 633.33 μs | 338.71 μs | Improved (-46.5%) |
| test_spline_1d_forward[batch_32-num_points_512]               |   2.40 ms | 860.23 μs | Improved (-64.2%) |
| test_spline_1d_forward[batch_128-num_points_128]              | 638.83 μs | 328.01 μs | Improved (-48.7%) |
| test_spline_1d_forward[batch_128-num_points_512]              |   2.76 ms |   1.07 ms | Improved (-61.2%) |
| test_spline_1d_forward[batch_512-num_points_128]              | 715.05 μs | 388.40 μs | Improved (-45.7%) |
| test_spline_1d_forward[batch_512-num_points_512]              |   4.42 ms |   1.54 ms | Improved (-65.1%) |
| test_single_qtransform_compute_qtiles[batch_32-sr_1024_q_12]  |   6.63 ms |   1.22 ms | Improved (-81.5%) |
| test_single_qtransform_compute_qtiles[batch_32-sr_4096_q_12]  |  13.28 ms |   1.98 ms | Improved (-85.1%) |
| test_single_qtransform_compute_qtiles[batch_32-sr_4096_q_64]  |  32.61 ms |   1.36 ms | Improved (-95.8%) |
| test_single_qtransform_compute_qtiles[batch_128-sr_1024_q_12] |   6.68 ms |   1.22 ms | Improved (-81.7%) |
| test_single_qtransform_compute_qtiles[batch_128-sr_4096_q_12] |  13.22 ms |   2.17 ms | Improved (-83.6%) |
| test_single_qtransform_compute_qtiles[batch_128-sr_4096_q_64] |  32.54 ms |   1.86 ms | Improved (-94.3%) |
[...]

PR #319, opened 2026-06-14
faster q-transform

mcoughlin:
This PR makes a few q-transform optimizations.


wbenoit26:
I actually made what I think is the equivalent change in #314 just a couple weeks ago. The changes live on the `dev` branch right now.

PR #331, opened 2026-06-25
Update email and maintainers

wbenoit26:
Adds Bhavya to the list of maintainers and post users to emailing one of the maintainers rather than the stale ml4gw email.


deepchatterjeeligo (approved):
Thanks.

PR #339, opened 2026-08-03
preliminary AI policy

deepchatterjeeligo:
I happened to land on Scipy's AI use policy. I think we should add a statement about the use of AI in proposing code changes. Please feel free to change the language.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,20 @@ if an automatic fix is not possible.
    motivated for gravitational-wave data analysis. For example, if you have used your 
    personal fork of `ml4gw` and performed an analysis that is published, it can be a strong
    usecase.
-4. Please [cite](/CITATION.cff) the repository if you have used `ml4gw` in your work.
+4. Please adhere to the [AI-policy](#ai-policy) below. 
+5. Please [cite](/CITATION.cff) the repository if you have used `ml4gw` in your work.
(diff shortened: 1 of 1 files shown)


wbenoit26:
This is a good idea - it's something I've been thinking about recently, and we could consider making it a ML4GW-level policy. I think the main point of friction will be in making sure contributors actually do have an understanding of what they submit, because that could still be a decent amount of work.

wbenoit26 (approved):
"Please disclose if you have made significant use of AI tools in proposing a patch." This will be the hardest thing for people to make a judgement call on. But I think this is a good start, and we can update the language if we need to. As we make PRs, we can try to set an example of what we're looking for.

bhgupta20 (approved):
I agree with both of you. This will be a good addition (and most likely needed in the coming times), and we can adopt it for the full ML4GW suite as Will mentioned. The wording of the statement looks good to me.

deepchatterjeeligo:
I agree that the phrasing is subjective w.r.t. the level of AI use. Hopefully, as time goes by the community will come up with better metrics. I skimmed some of the other links ([this one](https://blog.scientific-python.org/scientific-python/community-considerations-around-ai/) is a nice read, but a bit long) and find "understand before you create a PR" to be the main point. So I am hoping that we convey the same for now, although in a loose sense.

PR #343, opened 2026-08-23
test(nn): add unit tests for autoencoder library (#87)

### LINKED ISSUE #87: Tests for `autoencoder` library
Before merging `dev` into main, we should write some basic unit tests for the autoencoder library

lucacirfeta:
Resolves #87

### Summary
This PR adds unit test coverage for the `ml4gw.nn.autoencoder` subpackage, including base abstractions, convolutional autoencoders, skip connection modules, and resizing utilities.

> [!NOTE]
> **Note for reviewers:** While writing the tests, I encountered and fixed 4 runtime bugs in the underlying autoencoder code (detailed under *Small bugfixes* below). Because these modify runtime behavior rather than just adding test files, I wanted to flag them explicitly — in particular the `ConvolutionalAutoencoder` skip-connection adjustment.
> 
> If you prefer to keep this PR strictly test-only and handle the fixes in a dedicated PR (or discuss any architectural preference on the skip connection handling), I am very happy to split them.

### What is covered
- `ml4gw.nn.autoencoder.base.Autoencoder`:
  - Empty, single-block, and multi-block sequential pipelines.
  - Isolated calls to `encode` and `decode` (with and without `return_states`).
  - Validation of intermediate states required by skip connections (raising explicit `ValueError` when missing or mismatched).
- `ml4gw.nn.autoencoder.skip_connection`:
  - Output channel arithmetic (`get_out_channels`) for base, add, and concat skip connections.
  - Forward dimension matching and channel interleaving for grouped `ConcatSkipConnect`.
  - Explicit error handling when input channels cannot be divided evenly across groups.
- `ml4gw.nn.autoencoder.convolutional`:
  - `ConvBlock` and `ConvolutionalAutoencoder` parametrized across channels, kernel sizes (3, 5), strides (1, 2), groups, and custom output activations.
  - Exact output shape preservation across arbitrary input lengths via `match_size`.
  - Gradient backward pass test (`loss.backward()`) ensuring all trainable parameters receive finite, non-NaN gradients.
- `ml4gw.nn.autoencoder.utils.match_size`:
  - Unit tests covering exact match, zero-padding (odd/even), and symmetric cropping (odd/even).

### Small bugfixes revealed during testing
- Fixed slicing in `match_size` when `target_size < X.size(-1)`.
- Fixed intermediate state indexing in `Autoencoder.decode` (`states[i]` instead of double reverse index).
- Ensured uninstantiated `activation` classes passed to `ConvBlock` are initialized as callable modules.
- Adjusted skip connection attachment in `ConvolutionalAutoencoder` so the bottleneck layer decodes directly without expecting an upstream skip state.

### What is explicitly NOT covered
- 2D / 3D spatial autoencoders (currently only 1D conv architectures are defined in this module).
- Dynamic argument inspection between skip connection and autoencoder blocks (marked as TODO in source).

### Verification
Ran locally with `pytest`:
- `tests/nn/autoencoder/`: 166 passed.
- Full test suite: 1948 passed.
- Pre-commit formatting (`uv run prek run --all-files`): all checks passed.

### DIFF
--- a/ml4gw/nn/autoencoder/base.py
+++ b/ml4gw/nn/autoencoder/base.py
@@ -43,6 +43,9 @@ class Autoencoder(torch.nn.Module):
                 X = block.encode(X)
             states.append(X)
 
+        if len(self.blocks) == 0 and isinstance(X, tuple) and len(X) == 1:
+            X = X[0]
+
         # don't need to return the last
         # state, since that's just equal
         # to the output of this layer
@@ -72,9 +75,13 @@ class Autoencoder(torch.nn.Module):
             else:
                 X = block.decode(X)
 
-            state = states[-i - 1]
-            if state is not None:
-                X = self.skip_connection(X, state)
+            if states is not None:
+                state = states[i]
+                if state is not None:
+                    X = self.skip_connection(X, state)
+
+        if len(self.blocks) == 0 and isinstance(X, tuple) and len(X) == 1:
+            return X[0]
         return X
 
     def forward(self, *X: Tensor) -> Tensor:
--- a/ml4gw/nn/autoencoder/base.py
+++ b/ml4gw/nn/autoencoder/base.py
@@ -54,7 +54,12 @@ class Autoencoder(torch.nn.Module):
         return X
 
     def decode(self, *X, states: Sequence[Tensor] | None = None) -> Tensor:
-        if self.skip_connection is not None and states is None:
+        if self.skip_connection is None and states is not None:
+            raise ValueError(
+                "Cannot pass intermediate states when autoencoder "
+                "has no skip connection function specified"
+            )
+        elif self.skip_connection is not None and states is None:
             raise ValueError(
                 "Must pass intermediate states when autoencoder "
                 "has a skip connection function specified"
--- a/ml4gw/nn/autoencoder/convolutional.py
+++ b/ml4gw/nn/autoencoder/convolutional.py
@@ -55,11 +55,19 @@ class ConvBlock(Autoencoder):
             groups=groups,
         )
 
-        self.activation = activation()
+        self.activation = (
+            activation() if isinstance(activation, type) else activation
+        )
         if output_activation is not None:
-            self.output_activation = output_activation
+            self.output_activation = (
+                output_activation()
+                if isinstance(output_activation, type)
+                else output_activation
+            )
         else:
-            self.output_activation = activation
+            self.output_activation = (
+                activation() if isinstance(activation, type) else activation
+            )
 
         self.encode_norm = norm(out_channels)
         self.decode_norm = norm(decode_channels)
@@ -118,11 +126,10 @@ class ConvolutionalAutoencoder(Autoencoder):
             # case it will just be in_channels anyway)
             decode = in_channels if i else decode_channels
 
-            # don't have the middle layer skip to itself
-            # TODO: wait I don't think this makes sense.
-            # j = len(encode_channels) - 1 - i
-            # connect = skip_connection if j else None
-            connect = skip_connection
+            # Deepest block receives bottleneck encoding
+            # directly without skip connection
+            j = len(encode_channels) - 1 - i
+            connect = skip_connection if j else None
 
             # all intermediate layers should use the same
             # activation. Only the last decoder should have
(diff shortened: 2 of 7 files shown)


wbenoit26:
Hi @lucacirfeta, thanks for this PR and thanks for your patience as we've had people traveling/on vacation. I'll try to review this in the next day or two, though from a glance it looks like a useful addition.

tests/nn/autoencoder/test_autoencoder_utils.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+import torch
+
+from ml4gw.nn.autoencoder.utils import match_size
+
+
+@pytest.mark.parametrize("target_size", [8, 16, 23, 32])
+@pytest.mark.parametrize("input_size", [8, 15, 16, 27, 32])
+def test_match_size(target_size, input_size):
```
wbenoit26, line 9: This test re-implements the function itself, so it doesn't serve to check whether the function does what we want. Could you instead have the test locate the content of the smaller tensor within the larger tensor and test that `left` and `right` are within 1 sample of each other?

E.g., if the tensor is getting expanded, locate the original tensor within the padded version, measure its distance from each edge, and confirm that `abs(left - right) <= 1`.

tests/nn/autoencoder/test_base.py
```diff
@@ -0,0 +1,112 @@
...
+    # Forward
+    out = ae(x)
+    assert out.shape == x.shape
+
+
+def test_autoencoder_with_skip_connection():
+    skip = AddSkipConnect()
+    ae = Autoencoder(skip_connection=skip)
+    # Using blocks with equal in/out channels
+    # so AddSkipConnect can add x + state
```
wbenoit26: I suspect that the reason that the input/output channels need to be the same is the indexing bug that you fixed in `base.py`. With that fix in place, can you try uneven channels?

tests/nn/autoencoder/test_convolutional.py
```diff
@@ -0,0 +1,144 @@
...
+)
+def test_convolutional_autoencoder_shapes(
+    in_channels,
+    encode_channels,
+    kernel_size,
+    stride,
+    decode_channels,
+    skip_connection,
+):
+    # When using AddSkipConnect with ConvolutionalAutoencoder,
```
wbenoit26: Likewise here. I think these lines can be removed.

ml4gw/nn/autoencoder/convolutional.py
wbenoit26: We can simplify the changes below with the type definition from line 10. In the signature, we can set:
```python
activation: Module = torch.nn.ReLU,
...
output_activation: Module | None = None,
```

And then in the body:
```python
self.activation = activation()
self.output_activation = (output_activation or activation)()
```

All the checks can then be removed.

ml4gw/nn/autoencoder/base.py
wbenoit26: Pre-existing issue that I'm noticing on this review: we don't raise an error when `self.skip_connection` is `None`, but `states` is not `None`. Would you mind adding a check for that?

wbenoit26 (changes requested):
This looks great – thanks for adding tests and making the autoencoder module usable. There's a few things that I'd like to see changed in the code; I've commented on them below.

A couple other items that would be nice to have, but that I wouldn't hold up this PR over:
1. Benchmarks: In `benchmarks/nn` we have benchmarks against resnet, s4d, etc, but none for the autoencoder. Would be great to add some here as well.
2. Docstrings: Now that the code is functional, docstrings for `ConvBlock` and `match_size` would be very welcome.

Also, we [recently added](https://github.com/ML4GW/ml4gw/pull/339) an AI disclosure policy. I'm not concerned about the content of this PR, but for completeness, if you used AI, could you please add a statement?

lucacirfeta:
Thanks for review!

I addressed the requested changes:

- Reworked the `match_size` test to locate the original content and verify
  that the left/right padding or cropping differs by at most one sample.
- Tested skip connections with uneven channel sizes.
- Removed the channel-equality workaround from the convolutional tests.
- Simplified activation handling to use module constructors.
- Added an explicit error when intermediate states are passed without a
  configured skip connection.
- Added docstrings for `ConvBlock` and `match_size`.

Validation:
- `459 passed` in `tests/nn`

### AI disclosure

AI tools were used to assist with review analysis, implementation, and test
design. I reviewed the final changes, verified the resulting diff, and ran
the tests reported in this PR.

wbenoit26 (approved):

wbenoit26:
Great, thanks for making these updates.

PR #346, opened 2026-09-03
fix(slicing): handle full-length kernels (#86)

### LINKED ISSUE #86: `min_val == max_val` in `sample_kernels`
If the tensor you want to sample kernels from already has the length of the kernels you want to sample, `min_val` will equal `max_val` and `randint` will raise an error. In this case, it might be worth doing this check explicitly and either 1) returning the tensor as is or 2) raising an error.

lucacirfeta:
## Summary

- handle the single valid sampling position when the input length equals `kernel_size`
- preserve the existing output and index shapes for 1D, 2D coincident/non-coincident, and 3D inputs
- add regression coverage for each supported input shape

Closes #86

## Verification

- `.venv/Scripts/python.exe -m pytest tests/utils -q`: 128 passed
- `uvx prek run --all-files`: all hooks passed

## AI assistance

I used OpenAI Codex to help investigate, implement, and test this patch. I reviewed the diff and verified the behavior locally.

### DIFF
--- a/ml4gw/utils/slicing.py
+++ b/ml4gw/utils/slicing.py
@@ -313,11 +313,7 @@ def sample_kernels(
         )
 
     if X.ndim == 1:
-        max_idx = len(X) - kernel_size
-        if max_idx == 0:
-            idx = torch.zeros(N, dtype=torch.long)
-        else:
-            idx = torch.randint(max_idx, size=(N,))
+        idx = torch.randint(len(X) - kernel_size + 1, size=(N,))
         if return_idx:
             return slice_kernels(X, idx, kernel_size), idx
         return slice_kernels(X, idx, kernel_size)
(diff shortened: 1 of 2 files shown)


wbenoit26:
Thanks for the PR. The test looks good, but I think there's a better fix available for the issue. The problem is really that the upper bound for `torch.randint` is exclusive, so we were never actually allowing `max_idx` to be sampled. Instead of the checks, we can do `idx = torch.randint(len(X) - kernel_size + 1, size=(N,))` [here](https://github.com/ML4GW/ml4gw/blob/188b800b05a6ba5f28c7732c6146bf4b4d10bc9b/ml4gw/utils/slicing.py#L316) and `min_val, max_val = 0, X.shape[-1] - kernel_size + 1` [here](https://github.com/ML4GW/ml4gw/blob/188b800b05a6ba5f28c7732c6146bf4b4d10bc9b/ml4gw/utils/slicing.py#L325). There aren't any other paths by which `min_val == max_val` that aren't already handled.

lucacirfeta:
Hi, thanks for the review.
I replaced the special-case branches with the inclusive upper bound:

- `len(X) - kernel_size + 1` for 1D inputs
- `X.shape[-1] - kernel_size + 1` for unrestricted multidimensional sampling

This makes the final valid kernel position sampleable and naturally handles
full-length kernels. The tests were updated accordingly.

Validation:
- `128 passed` in `tests/utils`

wbenoit26 (approved):
Perfect, thank you for doing that.

PR #347, opened 2026-09-06
fix(waveforms): fix ringdown scaling and validate against LALSuite (#141)

### LINKED ISSUE #141: Add tests for the ringdown waveforms
I merged the ringdown waveform PR (#130) without tests in place because, at the end of the day, it's just a damped sine function and we can start meaningfully testing our ability to detect ringdowns even if we don't match LAL. However, before we do a serious search for ringdowns, we should get that matching to happen, and we should have tests in here that check for this.

lucacirfeta:
## Summary

- fix the ringdown amplitude to scale with the square root of the radiated mass fraction, `sqrt(epsilon)`
- use the standard `cos(inclination)` factor for the cross polarization
- add a batched comparison against `lalsimulation.SimBlackHoleRingdown` for dimensionless spins 0.0, 0.5, and 0.9

## Numerical validation

`Ringdown` uses closed-form fits for the black-hole mass, spin, and angular dependence, while LALSuite computes the Kerr modes numerically. The test therefore compares the waveforms using relative L2 error rather than requiring sample-exact equality.

Across the tested spin range, both polarizations remain within 7% of the LALSuite reference. Before the corrections, the new test failed with approximately 90% relative error.

## Testing

- `python -m pytest tests/waveforms/adhoc -q` — 49 passed
- `uvx prek run --files ml4gw/waveforms/adhoc/ringdown.py tests/waveforms/adhoc/test_ringdown.py` — all hooks passed

Closes #141

## AI disclosure

AI tools were used to assist with reference analysis, implementation, and test design. The final changes were validated against LALSuite, and the resulting diff and reported test results were reviewed before submission.

### DIFF
--- a/tests/waveforms/adhoc/test_ringdown.py
+++ b/tests/waveforms/adhoc/test_ringdown.py
@@ -1,0 +1,72 @@
+import lal
+import lalsimulation
+import numpy as np
+import torch
+
+from ml4gw.waveforms import Ringdown
+
+
+def test_ringdown_matches_lalsimulation():
+    sample_rate = 4096
+    duration = 1
+    mass = 100
+    spins = [0.0, 0.5, 0.9]
+    epsilon = 0.01
+    phase = 0.3
+    inclination = 0.7
+    distance = 100
+
+    frequency, quality = zip(
+        *[
+            lalsimulation.SimBlackHoleRingdownMode(
+                mass * lal.MTSUN_SI, spin, 2, 2, -2
+            )
+            for spin in spins
+        ],
+        strict=True,
+    )
+    parameters = [
+        frequency,
+        quality,
+        [epsilon] * len(spins),
+        [phase] * len(spins),
+        [inclination] * len(spins),
+        [distance] * len(spins),
+    ]
+    parameters = [torch.tensor(x, dtype=torch.float64) for x in parameters]
+
+    ringdown = Ringdown(sample_rate, duration)
+    cross, plus = ringdown(*parameters)
+
+    # SimBlackHoleRingdown uses an azimuthal phase for the m=2 mode,
+    # whereas Ringdown accepts the phase of the damped sinusoid directly.
+    lal_phase = (np.pi - phase) / 2
+    distance *= 1e6 * lal.PC_SI
+
+    for i, spin in enumerate(spins):
+        hplus, hcross = lalsimulation.SimBlackHoleRingdown(
+            lal.LIGOTimeGPS(0),
+            lal_phase,
+            1 / sample_rate,
+            mass * lal.MSUN_SI,
+            spin,
+            epsilon,
+            distance,
+            inclination,
+            2,
+            2,
+        )
+        expected_plus = np.asarray(hplus.data.data)
+        expected_cross = np.asarray(hcross.data.data)
+        n_samples = len(expected_plus)
+
+        # Ringdown uses closed-form fits for the mass, spin, and angular
+        # dependence rather than LALSuite's numerical Kerr solution. The
+        # approximation stays within 7% of LALSuite over this spin range.
+        for actual, expected in (
+            (plus[i, :n_samples].numpy(), expected_plus),
+            (cross[i, :n_samples].numpy(), expected_cross),
+        ):
+            difference = np.linalg.norm(actual - expected)
+            relative_error = difference / np.linalg.norm(expected)
+            assert relative_error < 0.07
--- a/tests/waveforms/adhoc/test_ringdown.py
+++ b/tests/waveforms/adhoc/test_ringdown.py
@@ -7,42 +7,82 @@ import torch
 from ml4gw.waveforms import Ringdown
 
 
+def get_expected_amplitude(frequency, quality, epsilon, distance):
+    spin = 1 - (2 / quality) ** (20 / 9)
+    mass = (
+        (1 / (2 * np.pi))
+        * (lal.C_SI**3 / (lal.G_SI * frequency))
+        * (1 - 0.63 * (2 / quality) ** (2 / 3))
+    )
+    f_quality = 1 + (7 / 24) / quality**2
+    g_spin = 1 - 0.63 * (1 - spin) ** (3 / 10)
+    amplitude = (
+        np.sqrt(5 * epsilon / 2)
+        * (lal.G_SI * mass / lal.C_SI**2)
+        * quality ** (-0.5)
+        * f_quality ** (-0.5)
+        * g_spin ** (-0.5)
+    )
+    return amplitude / (distance * 1e6 * lal.PC_SI)
+
+
+def get_lal_time_evolution(sample_rate, mass, spin, epsilon, phase, distance):
+    # At zero inclination only the positive-m component remains. LAL uses an
+    # azimuthal phase for its (2, 2) mode, while Ringdown accepts the phase of
+    # the damped sinusoid directly, so this maps between the two conventions.
+    lal_phase = (np.pi - phase) / 2
+    hplus, hcross = lalsimulation.SimBlackHoleRingdown(
+        lal.LIGOTimeGPS(0),
+        lal_phase,
+        1 / sample_rate,
+        mass * lal.MSUN_SI,
+        spin,
+        epsilon,
+        distance * 1e6 * lal.PC_SI,
+        0,
+        2,
+        2,
+    )
+    waveform = np.asarray(hplus.data.data) + 1j * np.asarray(hcross.data.data)
+    # Dividing by the initial complex value removes LALSuite's numerical
+    # amplitude and spheroidal-harmonic phase while retaining its evolution.
+    return waveform / waveform[0] * np.exp(1j * phase)
+
+
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/waveforms/adhoc/test_ringdown.py
```diff
@@ -0,0 +1,72 @@
...
+
+def test_ringdown_matches_lalsimulation():
+    sample_rate = 4096
+    duration = 1
+    mass = 100
+    spins = [0.0, 0.5, 0.9]
+    epsilon = 0.01
+    phase = 0.3
+    inclination = 0.7
+    distance = 100
```
deepchatterjeeligo: Would you be able to convert to a few more combinations (maybe two for each parameter)? Using `pytest.mark.parametrize` might be the cleanest.

deepchatterjeeligo (changes requested):
Hi @lucacirfeta thank you for catching this and adding the unit test against lalsimulation. This overall looks good. Maybe you can use a couple of more combinations (maybe two for each parameter) for comparing against lalsimulation. I have left a comment.

tests/waveforms/adhoc/test_ringdown.py
lucacirfeta: Done — I parametrized two waveform configurations covering different sample rates, durations, masses, epsilon values, phases, inclinations, and distances. Each configuration is evaluated at spins 0.0, 0.5, and 0.9, for six LALSuite comparisons in total. Let me know if needs others 👍

wbenoit26: I do think we'll need a denser parameter grid. I tried various inclination angles and found disagreement much larger than 7%. The worst case was fully edge-on, where the error was more than 30%. Seems to be worse for higher spins as well.

wbenoit26: It may be possible to divide out LAL's numerical solution and multiply by the `ml4gw` factors to make the comparison exact. In that case, we should be clear about what we're actually testing, and it would be good to add a note in the docstring of the class.

lucacirfeta: Thanks for catching this — you were right that the 7% bound was based on an insufficient inclination grid and should not have been generalized.
I reworked the test so it no longer claims end-to-end agreement with LALSimulation's numerical spheroidal-harmonic amplitude and angular dependence. It now normalizes a face-on LAL (2, 2) waveform to retain its time evolution, then applies ML4GW's closed-form amplitude and inclination factors. The comparison is exact up to a 1e-6 numerical tolerance.
The new grid covers spins 0.0, 0.5, 0.9, and 0.99, and seven inclinations from 0 to pi, including the fully edge-on case, for both polarizations and both waveform configurations.
I also added a note to the Ringdown class docstring explicitly documenting the model boundary. The full local test suite passes (1955 passed), along with all pre-commit checks.
AI disclosure: This follow-up was prepared with assistance from Codex and was reviewed and validated by me.

wbenoit26: Thanks for the update! I had been thinking more along the lines of dividing out just the angular component, and leaving the amplitude alone. As its currently written, it looks like the test is comparing the amplitude against a re-implementation of the same thing, and so should always pass. Can you try leaving the amplitude from LAL as-is? I'm not sure that they'll match exactly, but it should be independent of inclination, at least.

lucacirfeta: Thanks, that makes sense. I revised the test so that it now removes only LALSuite's numerical spheroidal angular response using `SimBlackHoleRingdownSpheroidalWaveFunction`; the physical LAL strain amplitude and time evolution are left untouched.

Since the absolute normalization is not expected to agree exactly, the test computes the ML4GW-to-LAL scale for each case and verifies that, for a given spin, it remains consistent across all seven inclinations, both polarizations, and both waveform configurations. The residual normalization is spin-dependent (approximately 0.774–0.806 over the tested grid), as expected from the different amplitude models. The edge-on zero of the cross polarization is checked separately to avoid a singular angular division.

I also verified that the test fails when either of the original regressions is restored: linear epsilon scaling or the sine inclination factor. The class docstring now describes the remaining spin-dependent normalization difference more precisely.

Let me know :)

Validation:
- targeted ringdown test: 4 passed
- waveform adhoc suite: 52 passed
- full test suite: 1957 passed
- all relevant prek hooks passed

AI disclosure: This follow-up was prepared with assistance from Codex. I reviewed the resulting implementation and validated it against LALSuite, including explicit regression-mutation checks.

wbenoit26: Hi @lucacirfeta, I don't think you pushed your latest commit

lucacirfeta: Ops, sorry!
Done

tests/waveforms/adhoc/test_ringdown.py
```diff
@@ -44,80 +27,107 @@ def get_lal_time_evolution(sample_rate, mass, spin, epsilon, phase, distance):
...
+                    continue
+
+                expected = angular_factor * waveform
+                scale = (
+                    np.vdot(expected, actual).real
+                    / np.vdot(expected, expected).real
+                )
+                assert scale > 0
+                residual = np.linalg.norm(actual - scale * expected)
+                residual /= np.linalg.norm(expected)
```
wbenoit26: I think this should be `np.linalg.norm(actual)`

tests/waveforms/adhoc/test_ringdown.py
```diff
@@ -44,80 +27,107 @@ def get_lal_time_evolution(sample_rate, mass, spin, epsilon, phase, distance):
...
+                assert residual < 1e-6, (
+                    f"{polarization} polarization shape differs for "
+                    f"spin={spin} and inclination={inclination}"
+                )
+                scale_factors.append(scale)
+
+    # The closed-form amplitude has a spin-dependent normalization relative to
+    # LALSuite, but its scaling must be independent of all other parameters,
+    # inclination, and polarization.
+    np.testing.assert_allclose(
```
wbenoit26, line 141: This is great; could you also add a test bounding `scale` for `spin=0` and the LAL waveform unmodified? At `spin=0`, the spherical and spheroidal harmonics should agree and the only difference should come from the numerical methods that LAL uses.

tests/waveforms/adhoc/test_ringdown.py
```diff
@@ -0,0 +1,133 @@
...
+    return waveform / angular_component
+
+
+@pytest.mark.parametrize("spin", [0.0, 0.5, 0.9, 0.99])
+def test_ringdown_matches_lal_scaling_after_angular_correction(spin):
+    configurations = [
+        (2048, 0.5, 50.0, 0.01, 0.0, 100.0),
+        (4096, 1.0, 100.0, 0.04, 0.3, 200.0),
+    ]
+    inclinations = np.linspace(0, np.pi, 7)
```
deepchatterjeeligo, line 73: What I meant was to use all of these parameters under the `pytest.mark.parametrize` decorator. My hope is that after this, you will not have a for loop like currently in line 50

deepchatterjeeligo: For example, if you look at the test [here](https://github.com/ML4GW/ml4gw/blob/main/tests/waveforms/cbc/test_cbc_waveforms.py#L95) you will find something similar. So I was expecting something like
```
@pytest.mark.parametrize("spin,sample_rate,duration,....", [[0.0, 0.5, 0.9, 0.99], [2048, 4096], [0.5, 1.0], ...])
```

tests/waveforms/adhoc/test_ringdown.py
lucacirfeta: Updated — the residual is now normalized by `np.linalg.norm(actual)`.

tests/waveforms/adhoc/test_ringdown.py
lucacirfeta: Added. The new test compares ML4GW directly against the unmodified LAL waveform at `spin=0`, for both waveform configurations and both polarizations. The measured scale is approximately 1.0173, so the test bounds it to within 2% of unity while separately checking the waveform residual.

tests/waveforms/adhoc/test_ringdown.py
lucacirfeta: Updated — the four spins and both complete waveform configurations are now expressed through `pytest.mark.parametrize`, producing eight independently reported test cases. The configuration loop has been removed while retaining the batched seven-inclination coverage.

Local validation:
- targeted ringdown tests: 10 passed
- waveform adhoc suite: 58 passed
- full suite: 1963 passed
- all relevant prek hooks passed

I also repeated the mutation checks: the tests reject both the previous linear-epsilon scaling and the sine inclination factor.

AI disclosure: This follow-up was prepared with assistance from Codex. I reviewed the implementation and validated the final diff against LALSuite and the complete local test suite.

wbenoit26 (commented):
Thanks @lucacirfeta, I'm happy with how this looks. @deepchatterjeeligo if you're also satisfied with the tests, I'm good to approve this and merge it in.

deepchatterjeeligo (approved):
LGTM thank you @lucacirfeta

PR #349, opened 2026-09-09
feat(transforms): add causal minimum-phase whitening (#246)

### LINKED ISSUE #246: Translate Zero-Latency Whitening to Torch
The [zlw](https://git.ligo.org/james.kennington/zlw) package has a nice implementation of a zero-latency whitening filter (as well as the standard linear-phase whitener). We should translate the functionality to `ml4gw` to improve our online latency.

lucacirfeta:
Closes #246.

## Summary

- add a Torch implementation of minimum-phase whitening filter construction using folded real-cepstrum spectral factorization
- add a fittable `MinimumPhaseWhiten` transform that applies the filter as a grouped causal FIR convolution
- preserve ML4GW's one-sided PSD normalization convention
- support both even- and odd-length impulse responses
- add scientific tests, usage documentation, and a GPU benchmark

## Motivation

The existing whitening transform relies on an acausal frequency-domain filter and therefore requires future samples and symmetric edge cropping.

The new transform fits a minimum-phase FIR kernel from background data or a PSD and applies it using only the current and previous samples. This provides a whitening path suitable for low-latency online inference.

The transform uses zero-valued history at the beginning of each input. Streaming callers can prepend `kernel_length * sample_rate - 1` historical samples and discard the corresponding warm-up outputs.

A running mean is deliberately not removed because that would introduce a future-sample dependency.

## Validation

- full ML4GW test suite: `1965 passed`
- targeted spectral and whitening tests passed on Windows and Linux/WSL
- `ruff`, `prek`, and `git diff --check` passed
- comparison with the reference `zlw` implementation:
  - maximum absolute NumPy difference: `4.8e-17`
  - exact agreement using the Torch backend
- analytical first-order filter validation
- end-to-end AR(1) colored-noise whitening validation
- explicit causality test confirming that an impulse produces no earlier output

## GPU benchmark

Environment:

- NVIDIA GeForce RTX 5070
- PyTorch 2.10.0 with CUDA 12.8

| Batch size | Mean |
| ---: | ---: |
| 32 | 0.262 ms |
| 128 | 0.676 ms |
| 512 | 2.340 ms |

All 9 whitening benchmark configurations passed.

## AI usage disclosure

OpenAI Codex was used substantially to assist with investigation, implementation, test design, documentation, and validation. Its output was checked against the reference `zlw` behavior, analytical test cases, and the complete ML4GW test suite.

### DIFF
--- a/benchmarks/transforms/test_whitening.py
+++ b/benchmarks/transforms/test_whitening.py
@@ -3,7 +3,7 @@
 import torch
 from constants import KERNEL_LEN, NUM_CHANNELS, NUM_SAMPLES, SAMPLE_RATE
 
-from ml4gw.transforms import FixedWhiten, Whiten
+from ml4gw.transforms import FixedWhiten, MinimumPhaseWhiten, Whiten
 
 FDURATION = 0.5
 NUM_SAMPLES_WHITEN = SAMPLE_RATE * 4
@@ -31,5 +31,21 @@ def test_fixed_whiten_forward(benchmark, batch_size, device, maybe_sync):
     bg = torch.randn(NUM_CHANNELS, NUM_SAMPLES)
     whitener.fit(FDURATION, bg[0], bg[1], fftlength=KERNEL_LEN)
     whitener = whitener.to(device)
+    x = torch.randn(batch_size, NUM_CHANNELS, NUM_SAMPLES, device=device)
+    benchmark(maybe_sync(whitener), x)
+
+
+def test_minimum_phase_whiten_forward(
+    benchmark, batch_size, device, maybe_sync
+):
+    whitener = MinimumPhaseWhiten(
+        num_channels=NUM_CHANNELS,
+        kernel_length=FDURATION,
+        sample_rate=SAMPLE_RATE,
+    )
+    num_freqs = int(FDURATION * SAMPLE_RATE) // 2 + 1
+    psd = torch.ones(num_freqs)
+    whitener.fit(*(psd for _ in range(NUM_CHANNELS)))
+    whitener = whitener.float().to(device)
     x = torch.randn(batch_size, NUM_CHANNELS, NUM_SAMPLES, device=device)
     benchmark(maybe_sync(whitener), x)
--- a/docs/examples/transforms.whitening.rst
+++ b/docs/examples/transforms.whitening.rst
@@ -27,3 +27,29 @@ Whitening
    # power spectral density computed using, e.g., the 
    # `SpectralDensity` transform.
    X_whitened = whitener(X, psd)
+
+Minimum-phase whitening
+-----------------------
+
+For online applications, a minimum-phase filter can whiten data without
+depending on future samples. Fit the filter from one background timeseries or
+PSD per channel, then apply it to tensors of any duration:
+
+.. code-block:: python
+
+   from ml4gw.transforms import MinimumPhaseWhiten
+
+   whitener = MinimumPhaseWhiten(
+      num_channels=2,
+      kernel_length=2,
+      sample_rate=2048,
+   )
+
+   # Passing fftlength means the inputs are interpreted as timeseries.
+   whitener.fit(background_h1, background_l1, fftlength=2)
+   X_whitened = whitener(X)
+
+The output has the same shape as the input. The first
+``kernel_length * sample_rate - 1`` samples use zero-valued history; when
+processing consecutive chunks, prepend that many real historical samples and
+discard their outputs.
--- a/docs/examples/transforms.whitening.rst
+++ b/docs/examples/transforms.whitening.rst
@@ -50,6 +50,7 @@ PSD per channel, then apply it to tensors of any duration:
    X_whitened = whitener(X)
 
 The output has the same shape as the input. The first
-``kernel_length * sample_rate - 1`` samples use zero-valued history; when
+``int(kernel_length * sample_rate) - 1`` samples use zero-valued history; when
 processing consecutive chunks, prepend that many real historical samples and
-discard their outputs.
+discard their outputs. Directly supplied PSDs at physical strain scale should
+use double precision to avoid underflow in ``torch.float32``.
--- a/ml4gw/spectral.py
+++ b/ml4gw/spectral.py
@@ -453,6 +453,8 @@ def minimum_phase_whitening_filter(
         psd:
             A strictly positive, finite one-sided power spectral density.
             Any leading dimensions are treated as batch dimensions.
+            Use double precision for unscaled physical PSDs whose values may
+            fall below the representable range of ``torch.float32``.
         n_fft:
             Length of the desired impulse response. If omitted, an even
             length of ``2 * (psd.size(-1) - 1)`` is inferred.
@@ -465,6 +467,13 @@ def minimum_phase_whitening_filter(
         ValueError:
             If the PSD shape is inconsistent with ``n_fft``, or if it
             contains a non-finite or non-positive value.
+
+    Notes:
+        The folded-cepstrum construction follows the
+        `zlw MPWhiteningFilter implementation`_ at commit ``b81871cd``.
+
+        .. _zlw MPWhiteningFilter implementation:
+            https://git.ligo.org/james.kennington/zlw/-/blob/b81871cdea5769bfb0a3b4ea57debbf6170d0c64/src/zlw/kernels.py
     """
     if psd.ndim == 0:
         raise ValueError("PSD must have a frequency dimension")
--- a/ml4gw/spectral.py
+++ b/ml4gw/spectral.py
@@ -439,6 +439,82 @@ def truncate_inverse_power_spectrum(
     return psd / 2
 
 
+def minimum_phase_whitening_filter(
+    psd: PSDTensor, n_fft: int | None = None
+) -> Tensor:
+    """Construct a causal, minimum-phase whitening filter.
+
+    The filter is obtained by spectral factorization of the inverse power
+    spectral density using a folded real cepstrum. Its one-sided frequency
+    response has magnitude ``1 / sqrt(psd)``, while its phase minimizes the
+    delay of the corresponding impulse response.
+
+    Args:
+        psd:
+            A strictly positive, finite one-sided power spectral density.
+            Any leading dimensions are treated as batch dimensions.
+        n_fft:
+            Length of the desired impulse response. If omitted, an even
+            length of ``2 * (psd.size(-1) - 1)`` is inferred.
+
+    Returns:
+        A real-valued impulse response with the same leading dimensions as
+            ``psd`` and length ``n_fft``.
+
+    Raises:
+        ValueError:
+            If the PSD shape is inconsistent with ``n_fft``, or if it
+            contains a non-finite or non-positive value.
+    """
+    if psd.ndim == 0:
+        raise ValueError("PSD must have a frequency dimension")
+    if psd.size(-1) < 2:
+        raise ValueError("PSD must contain at least two frequency bins")
+    if n_fft is None:
+        n_fft = 2 * (psd.size(-1) - 1)
+    if n_fft < 2 or psd.size(-1) != n_fft // 2 + 1:
+        raise ValueError(
+            "PSD length must equal n_fft // 2 + 1, but found "
+            f"{psd.size(-1)} bins for n_fft={n_fft}"
+        )
+    if not torch.is_floating_point(psd):
+        raise ValueError("PSD must be a floating-point tensor")
+    if not bool(torch.isfinite(psd).all()):
+        raise ValueError("PSD must contain only finite values")
+    if not bool((psd > 0).all()):
+        raise ValueError("PSD must contain only positive values")
+
+    log_amplitude = -0.5 * torch.log(psd)
+    reflection_end = -1 if n_fft % 2 == 0 else None
+    reflected = torch.flip(log_amplitude[..., 1:reflection_end], dims=(-1,))
+    log_amplitude = torch.cat((log_amplitude, reflected), dim=-1)
+
+    cepstrum = torch.fft.ifft(log_amplitude, dim=-1)
+    midpoint = n_fft // 2
+    if n_fft % 2 == 0:
+        folded = torch.cat(
+            (
+                cepstrum[..., :1],
+                2 * cepstrum[..., 1:midpoint],
+                cepstrum[..., midpoint : midpoint + 1],
+                torch.zeros_like(cepstrum[..., midpoint + 1 :]),
+            ),
+            dim=-1,
+        )
+    else:
+        folded = torch.cat(
+            (
+                cepstrum[..., :1],
+                2 * cepstrum[..., 1 : midpoint + 1],
+                torch.zeros_like(cepstrum[..., midpoint + 1 :]),
+            ),
+            dim=-1,
+        )
+
+    response = torch.exp(torch.fft.fft(folded, dim=-1))
+    return torch.fft.irfft(response[..., : psd.size(-1)], n=n_fft, dim=-1)
+
+
 def normalize_by_psd(
     X: WaveformTensor,
     psd: PSDTensor,
--- a/ml4gw/transforms/whitening.py
+++ b/ml4gw/transforms/whitening.py
@@ -291,9 +291,10 @@ class MinimumPhaseWhiten(FittableSpectralTransform):
     Unlike :class:`Whiten`, this transform never uses future samples to
     compute an output. The filter is fit once from a background PSD and then
     applied with left padding, which represents zero-valued history. Callers
-    processing consecutive chunks should prepend real input history and
-    discard the corresponding warm-up outputs. The transform does not remove
-    a running mean, since doing so would introduce a future-sample dependency.
+    processing consecutive chunks should prepend
+    ``int(kernel_length * sample_rate) - 1`` real input samples and discard
+    the same number of warm-up outputs. The transform does not remove a
+    running mean, since doing so would introduce a future-sample dependency.
 
     Args:
         num_channels:
@@ -342,7 +343,9 @@ class MinimumPhaseWhiten(FittableSpectralTransform):
             *background:
                 One time- or frequency-domain tensor per channel. Inputs are
                 interpreted as one-sided PSDs when ``fftlength`` is ``None``;
-                otherwise Welch PSDs are estimated from the timeseries.
+                otherwise Welch PSDs are estimated from the timeseries. Use
+                double precision for unscaled physical PSDs whose values may
+                fall below the representable range of ``torch.float32``.
             fftlength:
                 Length in seconds of Welch frames used for time-domain input.
             overlap:
--- a/ml4gw/transforms/whitening.py
+++ b/ml4gw/transforms/whitening.py
@@ -283,3 +283,111 @@ class FixedWhiten(FittableSpectralTransform):
         return spectral.normalize_by_psd(
             X, self.psd, self.sample_rate, pad, crop
         )
+
+
+class MinimumPhaseWhiten(FittableSpectralTransform):
+    """Whiten timeseries with a causal minimum-phase FIR filter.
+
+    Unlike :class:`Whiten`, this transform never uses future samples to
+    compute an output. The filter is fit once from a background PSD and then
+    applied with left padding, which represents zero-valued history. Callers
+    processing consecutive chunks should prepend real input history and
+    discard the corresponding warm-up outputs. The transform does not remove
+    a running mean, since doing so would introduce a future-sample dependency.
+
+    Args:
+        num_channels:
+            Number of channels to whiten.
+        kernel_length:
+            Length of the whitening filter in seconds.
+        sample_rate:
+            Sampling rate of input timeseries in Hz.
+        dtype:
+            Datatype used to store the fitted filter.
+
+    Shape:
+        - Input: ``(B, C, T)``
+        - Output: ``(B, C, T)``
+    """
+
+    def __init__(
+        self,
+        num_channels: int,
+        kernel_length: float,
+        sample_rate: float,
+        dtype: torch.dtype = torch.float64,
+    ) -> None:
+        super().__init__()
+        self.num_channels = num_channels
+        self.kernel_length = kernel_length
+        self.sample_rate = sample_rate
+
+        size = int(kernel_length * sample_rate)
+        if size < 2:
+            raise ValueError(
+                "Whitening filter must contain at least two samples"
+            )
+        kernel = torch.zeros((num_channels, 1, size), dtype=dtype)
+        self.register_buffer("kernel", kernel)
+
+    def fit(
+        self,
+        *background: TimeSeries1d | FrequencySeries1d,
+        fftlength: float | None = None,
+        overlap: float | None = None,
+    ) -> None:
+        """Fit a minimum-phase whitening filter to background data.
+
+        Args:
+            *background:
+                One time- or frequency-domain tensor per channel. Inputs are
+                interpreted as one-sided PSDs when ``fftlength`` is ``None``;
+                otherwise Welch PSDs are estimated from the timeseries.
+            fftlength:
+                Length in seconds of Welch frames used for time-domain input.
+            overlap:
+                Overlap in seconds between Welch frames. Defaults to half of
+                ``fftlength``.
+        """
+        if len(background) != self.num_channels:
+            raise ValueError(
+                f"Expected to fit whitening transform on {self.num_channels} "
+                f"background timeseries, but was passed {len(background)}"
+            )
+
+        num_freqs = self.kernel.size(-1) // 2 + 1
+        psds = [
+            self.normalize_psd(
+                x,
+                self.sample_rate,
+                num_freqs,
+                fftlength,
+                overlap,
+            )
+            for x in background
+        ]
+        psd = torch.stack(psds)
+        kernel = spectral.minimum_phase_whitening_filter(
+            psd, n_fft=self.kernel.size(-1)
+        )
+
+        # ML4GW PSDs are one-sided, so account for folded negative-frequency
+        # power and retain the unit-variance convention used by ``Whiten``.
+        kernel = kernel * (2 / self.sample_rate) ** 0.5
+        self.build(kernel=kernel[:, None])
+
+    def forward(self, X: TimeSeries3d) -> TimeSeries3d:
+        """Apply the fitted causal filter using zero-valued initial history."""
+        if X.ndim != 3 or X.size(1) != self.num_channels:
+            raise ValueError(
+                "Whitening transform expected input with shape "
+                f"(batch, {self.num_channels}, time), but found {X.shape}"
+            )
+
+        kernel = self.kernel.to(X)
+        X = torch.nn.functional.pad(X, (kernel.size(-1) - 1, 0))
+        return torch.nn.functional.conv1d(
+            X,
+            torch.flip(kernel, dims=(-1,)),
+            groups=self.num_channels,
+        )
--- a/tests/test_spectral.py
+++ b/tests/test_spectral.py
@@ -5,7 +5,81 @@ import pytest
 import torch
 from scipy import signal
 
-from ml4gw.spectral import fast_spectral_density, spectral_density, whiten
+from ml4gw.spectral import (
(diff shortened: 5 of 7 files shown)


wbenoit26:
Hi @lucacirfeta, thanks for creating this PR. I have a couple significant changes I want to request. I'll list out the major concerns, and also minor comments in-line.

- The reason we want the zero-latency whitening is for online, low-latency operations, and Aframe is the pipeline that we have running in low-latency. Aframe uses the `Whiten` transform, not `FixedWhiten`, for its training, testing, and real-time running. The existing class is fine as-is, but we'd really need an analogue of the `Whiten` class for this feature to be useful.
- The existing whitening transforms support bandpassing, which is critical for our applications. It's not obvious to me exactly how to implement bandpassing here, but it's important to figure out how to do so. Happy to iterate on this point.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,114 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+    applied with left padding, which represents zero-valued history. Callers
+    processing consecutive chunks should prepend
+    ``int(kernel_length * sample_rate) - 1`` real input samples and discard
+    the same number of warm-up outputs. The transform does not remove a
+    running mean, since doing so would introduce a future-sample dependency.
+
+    Args:
+        num_channels:
+            Number of channels to whiten.
+        kernel_length:
```
wbenoit26: Elsewhere, we use `fduration` for the filter length, and we should stick to that convention.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,114 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+        kernel = spectral.minimum_phase_whitening_filter(
+            psd, n_fft=self.kernel.size(-1)
+        )
+
+        # ML4GW PSDs are one-sided, so account for folded negative-frequency
+        # power and retain the unit-variance convention used by ``Whiten``.
+        kernel = kernel * (2 / self.sample_rate) ** 0.5
+        self.build(kernel=kernel[:, None])
+
+    def forward(self, X: TimeSeries3d) -> TimeSeries3d:
```
wbenoit26: The `forward` method for the other classes crops the corrupted parts of the input. The ZLW filter still corrupts some samples on the left side of the output, and we should crop that, and also have a `crop` parameter to allow a user to turn that off.

wbenoit26: Overall, it would be great to have this signature match the forward call of `FixedWhiten`

lucacirfeta:
Hi @wbenoit26 ,
thanks, both points make sense. I checked the current Aframe workflows and confirmed that they estimate a PSD per batch and call `Whiten(X, psd)`, so the fitted transform in this PR is not directly usable as a replacement.

To align the API with the existing transforms, I propose keeping the fitted implementation as `FixedMinimumPhaseWhiten` and adding a `MinimumPhaseWhiten` analogue of `Whiten` that accepts the PSD in `forward`, supports the same PSD broadcasting modes, and removes only the causal warm-up from the beginning of the output.

I would like to clarify the bandpass semantics before implementing them. The folded-cepstrum construction requires a strictly positive target magnitude, so directly applying the hard-zero frequency mask used by `Whiten` would introduce `log(0)`. The referenced `zlw` implementation also does not include bandpassing in its whitening kernel.

The two options I see are:

1. factorize a smoothly tapered response with a documented finite stopband floor; or
2. cascade a separate causal bandpass filter, which would introduce additional phase, state, and API requirements.

Would the first approach be appropriate? If so, is there a desired stopband attenuation or transition width, or should the goal be to approximate the effective response of the existing `Whiten` transform?

I also do not currently see the mentioned inline comments in the submitted review. Are those comments still pending?

ml4gw/spectral.py
```diff
@@ -439,6 +439,91 @@ def truncate_inverse_power_spectrum(
...
+        raise ValueError("PSD must contain only finite values")
+    if not bool((psd > 0).all()):
+        raise ValueError("PSD must contain only positive values")
+
+    log_amplitude = -0.5 * torch.log(psd)
+    reflection_end = -1 if n_fft % 2 == 0 else None
+    reflected = torch.flip(log_amplitude[..., 1:reflection_end], dims=(-1,))
+    log_amplitude = torch.cat((log_amplitude, reflected), dim=-1)
+
+    cepstrum = torch.fft.ifft(log_amplitude, dim=-1)
```
wbenoit26: I think we can use rfft/irffts here. All the input data should be real-valued.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,114 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+        """
+        if len(background) != self.num_channels:
+            raise ValueError(
+                f"Expected to fit whitening transform on {self.num_channels} "
+                f"background timeseries, but was passed {len(background)}"
+            )
+
+        num_freqs = self.kernel.size(-1) // 2 + 1
+        psds = [
+            self.normalize_psd(
```
wbenoit26: I think we'll run into issues with the PSD being interpolated to match the filter size. From what I recall from messing around with ZLW in the past is that you want to oversample the PSD, then truncate and taper.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,114 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+        """Apply the fitted causal filter using zero-valued initial history."""
+        if X.ndim != 3 or X.size(1) != self.num_channels:
+            raise ValueError(
+                "Whitening transform expected input with shape "
+                f"(batch, {self.num_channels}, time), but found {X.shape}"
+            )
+
+        kernel = self.kernel.to(X)
+        X = torch.nn.functional.pad(X, (kernel.size(-1) - 1, 0))
+        return torch.nn.functional.conv1d(
```
wbenoit26: Can we do the filtering in the frequency domain? Should be a decent bit faster.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,114 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+
+    def forward(self, X: TimeSeries3d) -> TimeSeries3d:
+        """Apply the fitted causal filter using zero-valued initial history."""
+        if X.ndim != 3 or X.size(1) != self.num_channels:
+            raise ValueError(
+                "Whitening transform expected input with shape "
+                f"(batch, {self.num_channels}, time), but found {X.shape}"
+            )
+
+        kernel = self.kernel.to(X)
```
wbenoit26: If a user passes `X` as float32, then this will convert any float64 background that was passed.

wbenoit26 (changes requested):
@lucacirfeta Here are the in-line comments. I need to look into how to approach the bandpassing and how to test whether whatever approach we go with is effective. I agree with your proposal for the API.

wbenoit26:
@lucacirfeta For the bandpassing, can you try using [`truncate_inverse_power_spectrum`](https://github.com/ML4GW/ml4gw/blob/7d7157fd6014c875440696a3ba0f0fbfc06ccc47/ml4gw/spectral.py#L348) directly to match what the current whitening modules do? I'm not certain what impact this will have on the minimum phase whitener itself.

lucacirfeta:
@wbenoit26 
Thanks, I tried this approach and it works well. The PSD is now passed through truncate_inverse_power_spectrum, including the requested highpass and lowpass, before minimum-phase factorization. This treats the existing truncated response as the target magnitude and only changes its phase to obtain a causal filter.
I added tests for unfiltered, highpass, lowpass, and bandpass configurations. The minimum-phase magnitude matches the standard truncated response within an absolute tolerance of 3e-3; away from the transition regions, the 95th-percentile passband deviation is below 1% and the stopband response is below 1e-3. A resolved narrow-line test also confirms that the factorized filter matches the standard truncated target.

ml4gw/transforms/whitening.py
lucacirfeta: Agreed. I renamed kernel_length to fduration throughout the minimum-phase transforms, tests, documentation, and benchmarks.

ml4gw/spectral.py
lucacirfeta: Agreed. The cepstral construction now uses irfft directly on the one-sided real log-amplitude spectrum and rfft after folding the real cepstrum. The tests cover both even and odd FFT lengths, as well as the expected inverse-ASD magnitude.

ml4gw/transforms/whitening.py
lucacirfeta: Agreed. The PSD is no longer downsampled to the final FIR length before factorization. Its native frequency grid is preserved, or upsampled only when it is shorter than twice the requested filter duration. truncate_inverse_power_spectrum is then applied before minimum-phase factorization.
I added tests covering the short-PSD upsampling path and confirming that a resolved narrow spectral feature reaches the factorization at its original resolution.

ml4gw/transforms/whitening.py
lucacirfeta: Updated. Filtering now uses zero-padded linear FFT convolution. The FFT length is the next power of two at or above T + L - 1, preventing circular aliasing, and the causal result is sliced back to the input duration before optional warm-up cropping.
A regression test verifies equivalence with grouped direct convolution. On an RTX 5070, the fixed transform improved by approximately 44%, 79%, and 92% for batch sizes 32, 128, and 512 respectively.

ml4gw/transforms/whitening.py
lucacirfeta: Fixed. The input and kernel are now promoted to their common dtype instead of converting the kernel to the input dtype. A fitted float64 kernel therefore remains float64 when the input is float32, and the output uses the promoted precision. This behavior is covered by a regression test.

ml4gw/spectral.py
lucacirfeta: Agreed. Both transforms now expose a crop: bool = True parameter. Since the FIR is causal, cropping removes the first fduration * sample_rate - 1 warm-up samples from the left. Passing crop=False preserves the input length and includes samples computed from zero-valued initial history.
I also split the API as discussed: MinimumPhaseWhiten.forward(X, psd, crop=True) accepts dynamic PSDs, while FixedMinimumPhaseWhiten.forward(X, crop=True) reuses a fitted filter. The dynamic transform supports PSD shapes (F,), (C, F), and (B, C, F).

ml4gw/transforms/whitening.py
lucacirfeta: Agreed. Both transforms now expose a crop: bool = True parameter. Since the FIR is causal, cropping removes the first fduration * sample_rate - 1 warm-up samples from the left. Passing crop=False preserves the input length and includes samples computed from zero-valued initial history.
I also split the API as discussed: MinimumPhaseWhiten.forward(X, psd, crop=True) accepts dynamic PSDs, while FixedMinimumPhaseWhiten.forward(X, crop=True) reuses a fitted filter. The dynamic transform supports PSD shapes (F,), (C, F), and (B, C, F).

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,313 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+        kernel = kernel.unsqueeze(0)
+
+    convolution_size = input_size + size - 1
+    n_fft = 1 << (convolution_size - 1).bit_length()
+    X_tilde = torch.fft.rfft(X, n=n_fft, dim=-1)
+    kernel_tilde = torch.fft.rfft(kernel, n=n_fft, dim=-1)
+    X = torch.fft.irfft(X_tilde * kernel_tilde, n=n_fft, dim=-1)
+    X = X[..., :input_size]
+    if crop:
+        X = X[..., size - 1 :]
```
wbenoit26: Let's crop `size` samples rather than `size - 1` to be consistent which the number of samples cropped from the `Whiten` class.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,313 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+    input_size = X.size(-1)
+    if crop and input_size < size:
+        raise ValueError(
+            f"Not enough timeseries samples {X.size(-1)} for number of "
+            f"warm-up samples {size - 1}"
+        )
+
+    dtype = torch.promote_types(X.dtype, kernel.dtype)
+    X = X.to(dtype=dtype)
+    kernel = kernel.to(device=X.device, dtype=dtype)
```
wbenoit26, line 387: The existing modules always return `torch.float32`. We probably shouldn't be forcing that, but for now, can you have your modules return in `float32` as well to match?

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,313 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+    The PSD is first processed by
+    :func:`~ml4gw.spectral.truncate_inverse_power_spectrum`, including any
+    requested highpass or lowpass response, then spectrally factorized into a
+    causal FIR filter.
+    """
+    if psd.size(-1) < 2:
+        raise ValueError("PSD must contain at least two frequency bins")
+    if not torch.is_floating_point(psd):
+        raise ValueError("PSD must be a floating-point tensor")
+    if not bool(torch.isfinite(psd).all()):
```
wbenoit26: This and the following check both call `bool` on a tensor, which would be expensive if the tensors are on GPU and this is happening on every forward call of `MinimumPhaseWhiten`. Let's delete these two checks from this function, they're duplicated in `spectral.minimum_phase_whitening_filter`.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,313 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+    while psd.ndim < 3:
+        psd = psd[None]
+    psd = spectral.truncate_inverse_power_spectrum(
+        psd,
+        size / sample_rate,
+        sample_rate,
+        highpass,
+        lowpass,
+    )
+    kernel = spectral.minimum_phase_whitening_filter(psd)[..., :size]
```
wbenoit26: Though this function also performs the checks on every forward call. We could put the checks in `spectral.minimum_phase_whitening_filter` behind a guard, like `if validate: ...` and then set `validate=False` in the call here.

benchmarks/transforms/test_whitening.py
```diff
@@ -33,3 +38,29 @@ def test_fixed_whiten_forward(benchmark, batch_size, device, maybe_sync):
...
     x = torch.randn(batch_size, NUM_CHANNELS, NUM_SAMPLES, device=device)
     benchmark(maybe_sync(whitener), x)
+
+
+def test_minimum_phase_whiten_forward(
+    benchmark, batch_size, device, maybe_sync
+):
+    whitener = MinimumPhaseWhiten(FDURATION, SAMPLE_RATE).to(device)
+    num_freqs = NUM_SAMPLES // 2 + 1
+    psd = torch.ones(NUM_CHANNELS, num_freqs, device=device)
```
wbenoit26: Can you copy the inputs from `Whiten` and `FixedWhiten` for the PSDs and `x` here? And then also compare benchmarks for both the original modules and the minimum phase modules? I'm curious how much slower the MP whitening is.

tests/transforms/test_whitening.py
```diff
@@ -204,3 +210,395 @@ def test_io(self, transform, background, tmp_path):
...
+class MinimumPhaseWhitenTest:
+    sample_rate = 256
+    fduration = 1
+    num_channels = 2
+
+    @property
+    def size(self):
+        return int(self.fduration * self.sample_rate)
+
+    def get_psds(self):
```
wbenoit26, line 226: Would be great to also have a realistic PSD at 2048 Hz

docs/examples/transforms.whitening.rst
```diff
@@ -27,3 +27,56 @@ Whitening
...
+   # Passing fftlength means the inputs are interpreted as timeseries.
+   whitener.fit(
+      background_h1,
+      background_l1,
+      fftlength=2,
+      highpass=20,
+   )
+   X_whitened = whitener(X)
+
+The highpass and lowpass responses use the same inverse-spectrum truncation
```
wbenoit26, line 70: I was looking into the effects of applying a bandpass after a minimum-phase whitening, and it looks like doing so introduces some group delay to the signal. This is going beyond the area that I'm familiar with, but we should document that impact.

docs/examples/transforms.whitening.rst
```diff
@@ -27,3 +27,56 @@ Whitening
...
+as the standard whitening transforms before the response is converted to a
+causal minimum-phase filter.
+
+By default, both transforms crop
+``int(fduration * sample_rate) - 1`` warm-up samples from the left edge. Pass
```
wbenoit26: We should also point out the difference in output alignment post-cropping between the two whitening filters.

wbenoit26 (changes requested):
Thanks for the updates @lucacirfeta, this looks quite good now. I've added a few more comments. I still need to take a thorough look through the tests, but I wanted to give you this feedback now.

ml4gw/transforms/whitening.py
lucacirfeta: Thanks. One alignment detail before I update the cropping: Whiten removes size // 2 samples from each edge, so the total is 2 * (size // 2)—size for even filter lengths but size - 1 for odd lengths. The minimum-phase implementation currently supports odd filter lengths as well. Would you prefer it to always crop size samples, or to crop 2 * (size // 2) so its output length matches Whiten for both even and odd lengths? For the even filter lengths used by Aframe, the two choices are identical.

wbenoit26: Let's make these match the existing modules so that they can be dropped into existing code without any modification. A future PR can allow both to handle even and odd filter lengths.

lucacirfeta: Thanks for clarifying. I now crop size samples as requested so that the minimum-phase transforms match the output length expected by existing code. I also updated the crop and streaming tests and documented that the causal transforms crop entirely from the left. I left harmonizing even and odd filter lengths across all whitening modules to a future PR, as suggested.

ml4gw/transforms/whitening.py
lucacirfeta: Updated. Both minimum-phase transforms now return torch.float32, including when crop=False, to match the existing whitening modules. The convolution still promotes the input and kernel to a common internal dtype, so a fitted float64 kernel is not reduced to float32 before the FFT computation. I added a test covering both the internal precision and the returned dtype.

ml4gw/transforms/whitening.py
lucacirfeta: Agreed. I removed the finite and positive tensor-value checks from _minimum_phase_kernel, so the dynamic forward path no longer performs those GPU-to-host synchronizations there. The inexpensive shape and dtype checks remain.

ml4gw/transforms/whitening.py
lucacirfeta: Implemented. spectral.minimum_phase_whitening_filter now has a validate=True argument, preserving validation for direct public calls. The internal whitening path calls it with validate=False after constructing the truncated inverse-spectrum response. Existing tests still cover the default validation behavior, and a new test verifies that the value validation is skipped when explicitly disabled.

benchmarks/transforms/test_whitening.py
lucacirfeta: Updated the benchmarks so that MinimumPhaseWhiten and FixedMinimumPhaseWhiten use the same input construction, dimensions, PSD shapes, and batch sizes as Whiten and FixedWhiten.
On my local RTX 5070, the median GPU timings were:
- Dynamic, batch 32: Whiten 577 µs, minimum phase 631 µs — about 9% slower.
- Dynamic, batch 128: Whiten 1131 µs, minimum phase 771 µs — about 32% faster.
- Dynamic, batch 512: Whiten 5074 µs, minimum phase 3804 µs — about 25% faster.
- Fixed, batch 32: FixedWhiten 222 µs, minimum phase 315 µs — about 42% slower.
- Fixed, batch 128: FixedWhiten 366 µs, minimum phase 631 µs — about 73% slower.
- Fixed, batch 512: FixedWhiten 945 µs, minimum phase 2054 µs — about 117% slower.
These results suggest that the dynamic implementation benefits from larger batches, while the fixed minimum-phase implementation is currently slower than FixedWhiten. I have not presented performance as a general advantage of the new transforms.

tests/transforms/test_whitening.py
lucacirfeta: Added. The test uses the LAL SimNoisePSDaLIGOaLIGO140MpcT1800545 PSD at 2048 Hz with a two-second filter and highpass/lowpass cutoffs. It verifies a finite float32 kernel and compares its magnitude response against the standard truncated inverse-spectrum target away from the transition regions. The relative-error checks cover the median, 95th percentile, and maximum passband deviation.

docs/examples/transforms.whitening.rst
lucacirfeta: Documented. The whitening documentation now explains that the causal minimum-phase response has frequency-dependent group delay and that changing the target magnitude with highpass or lowpass cutoffs also changes that delay, particularly near the transition frequencies. It also clarifies that cropping removes the initial filter warm-up but does not compensate for the group delay.

docs/examples/transforms.whitening.rst
lucacirfeta: Added. The documentation now explains that the existing whitening transform crops half of the filter duration from each edge, while the causal minimum-phase transform removes the full duration from the left. Therefore, even when the output lengths match, the two outputs should not be treated as sample-aligned.

tests/transforms/test_whitening.py
```diff
@@ -204,3 +212,438 @@ def test_io(self, transform, background, tmp_path):
...
+            response = 1 - coefficient * torch.exp(-1j * omega)
+            psds.append((2 / self.sample_rate) / response.abs().square())
+        return psds
+
+
+class TestMinimumPhaseWhiten(MinimumPhaseWhitenTest):
+    def get_transform(self):
+        return MinimumPhaseWhiten(self.fduration, self.sample_rate)
+
+    def test_crop_is_causal(self):
```
wbenoit26: Can you change the name to something like `test_crop_removes_settle_in`? This isn't actually testing causality.

tests/transforms/test_whitening.py
```diff
@@ -204,3 +212,438 @@ def test_io(self, transform, background, tmp_path):
...
+        torch.testing.assert_close(whitened, uncropped[..., self.size :])
+
+    @pytest.mark.parametrize("psd_ndim", [2, 3])
+    def test_psd_broadcasting(self, psd_ndim):
+        transform = self.get_transform()
+        X = torch.randn(4, self.num_channels, 1024)
+        psd = torch.full((2 * self.size + 1,), 2 / self.sample_rate)
+        expected = transform(X, psd, crop=False)
+
+        psd = psd.repeat(self.num_channels, 1)
```
wbenoit26: Because the PSD is being repeated, I don't think this sufficiently tests PSD broadcasting.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,314 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+                self.normalize_psd(
+                    x,
+                    self.sample_rate,
+                    num_freqs,
+                    fftlength,
+                    overlap,
+                )
+            )
+        psd = torch.stack(psds)
+        kernel = _minimum_phase_kernel(
```
wbenoit26, line 572: Given that we just `fit` once, we could call `spectral.minimum_phase_whitening_filter` with `validate=True`. It would just involve threading the `validate` parameter through `_minimum_phase_kernel`.

tests/transforms/test_whitening.py
```diff
@@ -204,3 +212,438 @@ def test_io(self, transform, background, tmp_path):
...
+            transform(X, torch.ones(2 * self.size + 1, dtype=torch.int64))
+
+    def test_upsamples_short_psd_before_factorization(self):
+        transform = self.get_transform()
+        psd = torch.full((self.size // 2 + 1,), 2 / self.sample_rate)
+        X = torch.randn(2, self.num_channels, 1024)
+
+        whitened = transform(X, psd, crop=False)
+
+        assert whitened.shape == X.shape
```
wbenoit26, line 431: This isn't testing whether the PSD was upsampled. If you do the same `patch` you did for `test_fit_preserves_resolved_psd_grid` below, you could directly test if the PSD had the expected number of samples.

ml4gw/transforms/whitening.py
```diff
@@ -283,3 +283,314 @@ def forward(self, X: TimeSeries3d, crop: bool = True) -> TimeSeries3d:
...
+
+def _minimum_phase_filter(
+    X: TimeSeries3d,
+    kernel: torch.Tensor,
+    crop: bool,
+) -> TimeSeries3d:
+    """Apply a causal FIR as a linear convolution in the frequency domain."""
+    size = kernel.size(-1)
+    input_size = X.size(-1)
+    if crop and input_size < size:
```
wbenoit26: Let's make this `<=`. As it is, we could return an empty tensor.

tests/test_spectral.py
```diff
@@ -1,11 +1,96 @@
...
+            minimum_phase_whitening_filter(psd)
+
+    def test_inconsistent_fft_length(self):
+        with pytest.raises(ValueError, match="PSD length"):
+            minimum_phase_whitening_filter(torch.ones(3), n_fft=8)
+
+    def test_skip_value_validation(self):
+        psd = torch.linspace(0.5, 4, 33, dtype=torch.float64)
+
+        with patch("ml4gw.spectral.torch.isfinite") as isfinite:
```
wbenoit26: This doesn't check whether the positivity test was skipped. You could instead pass a PSD with a zero bin and `validate=False` to check that no errors were raised.

wbenoit26 (changes requested):
I've taken a look through the tests. Just a few more changes, and I think we'll be good to merge this.

tests/transforms/test_whitening.py
lucacirfeta: Renamed to test_crop_removes_settle_in. The test only verifies that cropping removes the initial settle-in samples; causality remains covered separately by test_filter_is_causal.

tests/transforms/test_whitening.py
lucacirfeta: Updated. The broadcasting test now uses different PSDs for each channel and additional distinct scaling for each batch. It compares the broadcast result against independent one-dimensional PSD calls for every batch and channel, so incorrect channel or batch mapping can no longer pass because of repeated identical values.

ml4gw/transforms/whitening.py
lucacirfeta: Implemented. I threaded the validate argument through _minimum_phase_kernel. Dynamic MinimumPhaseWhiten calls retain validate=False to avoid synchronization on every forward call, while FixedMinimumPhaseWhiten.fit uses validate=True because factorization only happens once. The tests verify both paths explicitly.

tests/transforms/test_whitening.py
lucacirfeta: Updated. The test now patches minimum_phase_whitening_filter and directly verifies that the PSD passed to factorization contains size + 1 frequency bins after upsampling. It also checks that the dynamic path uses validate=False.

ml4gw/transforms/whitening.py
lucacirfeta: Changed to input_size <= size. I also added coverage for the exact boundary where the input length equals the filter length, preventing the cropped path from returning an empty tensor.

tests/test_spectral.py
lucacirfeta: Updated. The test now passes a PSD containing a zero bin with validate=False and verifies that factorization returns without raising a validation error. This directly exercises skipping the positivity check instead of checking only whether torch.isfinite was called.

wbenoit26 (approved):
Great, thanks @lucacirfeta this looks good to me now.

PR #358, opened 2026-09-16
Sync dev with main

wbenoit26:
Automated merge of `main` into `dev`.

Keeps `dev` mergeable so the eventual release merge stays small.
Review the diff, then use a **merge commit**; do not squash or rebase,
or the next sync will replay the same commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 ml4gw.egg-info
 docs/_build
 docs/jupyter_execute
+.coverage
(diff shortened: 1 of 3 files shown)


wbenoit26:
@bhgupta20 This PR contains the dependabot updates to main, plus the ZLW with its API converted to use ASD like the rest of the `dev` branch. Doesn't need a thorough review, because all the changes were reviewed previously.

PR #359, opened 2026-09-16
Docker

mcoughlin:
This PR creates a Dockerfile for container-based deployment.

Validated with Claude.


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,62 @@
...
+            type=sha
+
+      - uses: docker/setup-buildx-action@v3
+
+      # Push everywhere except pull requests (which build-only to catch breakage).
+      - name: Log in to Docker Hub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
wbenoit26, line 51: I don't think either of these secrets exist currently.

wbenoit26:
Other than needing to add the secrets, this looks fine to me. What's the use case?

wbenoit26:
Also, the docker workflow seems to be hanging - don't think it should take this long.

.github/workflows/docker.yaml
mcoughlin: Yes they will need to be added.

mcoughlin:
@wbenoit26 ability to run OSG-based analyses.

mcoughlin:
@wbenoit26 Caching is going to help a lot here after the first PR goes in.

wbenoit26:
I think this would actually make more sense as a PR against buoy itself. buoy depends on ml4gw, and it's where most of the build time is coming from. And then the version wouldn't need to be updated in the dockerfile whenever it changed. I'm happy to make the PR against buoy – any objections?

mcoughlin:
@wbenoit26 Whatever you think is best! Ultimately this container will be served over cvmfs, so it's worth thinking about the name a bit as well, eg to make it easy to identify

wbenoit26:
@mcoughlin I forgot I already had a workflow in buoy to build a container. You can do `apptainer pull buoy.sif docker://ghcr.io/ml4gw/buoy/buoy:main` to get the latest, or point to a particular version. Let me know if that works for what you need.

PR #362, opened 2026-09-20
Correct location of h5py dep

wbenoit26:
The h5py dependency was listed under the `dev` group, but it's imported in our dataloaders.


bhgupta20 (approved):
Agreed, good catch!

